PR #6, opened 2021-07-27
Alternate object storage support

hardiesoft:
Initial support for backblaze as an alternate s3 provider, and fixing up script locations in cron-jobs


CameronRP (approved):
Looks good.
Just wondering if we want to be calling it `minio` and `backblaze` and instead call it something more to do with it's purpose like `s3local` and `s3archive` as it doesn't actually have to be minio or backblaze.

hardiesoft:
Closed because it was merged into another PR

PR #7, opened 2021-07-28
Add device tests

gw7nvw:
Four changes:
- Remove the now-obsolete /api/test-cypress directory tree (these tests are now in /integration-tests, created when the repositories were merged into cacophony-web)
- Switch travis config to use the new /integration-tests test repository for build tests
- Add tests for "/devices" API endpoints in /integration-tests/cypress/integration/api/device* and supporting helpers in /integration-tests/cypress/commands/device*
- Restructure existing /integration-tests/cypress/commands API wrappers so that each API wrapper is found in in file named for the root-level API endpoint it accesses (previously seem to have been grouped non-thematically based on when they were written)


integration-tests/cypress/commands/api/device.d.ts
```diff
@@ -15,14 +15,31 @@ declare namespace Cypress {
...
-      log?: boolean
+      saltId: number,
+      log: boolean,
+      statusCode: number
+    );
+
+     /**
+     *register a device under a new group or name
+     *optionally check for an error response (statusCode!=200OK)
+     *optionally supply a password (autogenerat if not)
```
gferraro: autogenerate

integration-tests/cypress/commands/api/device.d.ts
```diff
@@ -34,12 +51,23 @@ declare namespace Cypress {
...
 
+
     /**
     * Retrieve device details using name and groupname
+    * use groupId if provided, otherwise groupName
     * compare with expected device details (JSON equivalent to that retunred by API)
     * optioanlly check for a non-200 status code
     */
-    apiCheckDevice(userName: string, cameraName: string, groupName: string, expectedDevice: ComparableDevice, statusCode: number): Chainable<Element>;
+    apiCheckDeviceInGroup(userName: string, cameraName: string, groupName: string, groupId: number, expectedDevice: all, statusCode: number);
```
gferraro: Should we stick to using device instead of camera?

integration-tests/cypress/commands/api/device.ts
```diff
@@ -0,0 +1,539 @@
...
+      {
+        camera: cameraName,
+        group: group,
+	saltId: saltId
+      },
+      log
+    );
+
+    const request = createCameraDetails(cameraName, group, null, saltId);
+    if(statusCode==200) {
```
gferraro, line 30: is this meant to check it is the supplied statusCode value?

integration-tests/cypress/commands/api/device.ts
```diff
@@ -0,0 +1,539 @@
...
+    } else {
+      checkRequestFails(request);
+    };
+  }
+);
+
+Cypress.Commands.add(
+  "apiDeviceReregister",
+  (oldName: string, newName: string, newGroup: string, password: string = null, keepName:boolean = false, statusCode: number = 200) => {
+    var uniqueName;
```
gferraro: maybe ask Jon but i think using let for variables is preferable (even if it is the same outcome)

integration-tests/cypress/commands/api/device.ts
```diff
@@ -0,0 +1,539 @@
...
+      `Reregister camera '${newName}' in group '${newGroup}'`,
+      {
+        camera: newName,
+        group: newGroup
+      },
+      true
+    );
+
+    if(password==null) {
+      password="p"+getTestName(newName);
```
gferraro: should this be based of uniqueName?

integration-tests/cypress/commands/api/device.ts
```diff
@@ -0,0 +1,539 @@
...
+      });
+    } else {
+      checkRequestFails(request);
+    };
+  }
+);
+
+Cypress.Commands.add(
+  "apiDeviceReregister",
+  (oldName: string, newName: string, newGroup: string, password: string = null, keepName:boolean = false, statusCode: number = 200) => {
```
gferraro: should new name be optional and remove keepName? and use that to base the logic off?

integration-tests/cypress/commands/api/device.ts
```diff
@@ -0,0 +1,539 @@
...
+      {
+        method: "POST",
+        url: fullUrl,
+        body: data
+      },
+      oldName,
+      statusCode
+    ).then((response) => {
+      const id = response.body.id;
+      saveCreds(response, newName, id);
```
gferraro, line 82: again uniqueName?

integration-tests/cypress/commands/api/device.ts
```diff
@@ -0,0 +1,539 @@
...
+    });
+  }
+);
+
+
+Cypress.Commands.add(
```
gferraro: I think this was already there, but it seems to be strange that it's a single command feels like it should be tested in a test

integration-tests/cypress/commands/api/device.ts
```diff
@@ -0,0 +1,539 @@
...
+      },
+      log
+    );
+
+    const request = createCameraDetails(cameraName, group, password, null, makeCameraATestName);
+    checkRequestFails(request);
+  }
+);
+
+function createCameraDetails(
```
gferraro: `createCamera` or better yet `createDevice`

integration-tests/cypress/commands/api/device.ts
```diff
@@ -0,0 +1,539 @@
...
+       if(statusCode==null || statusCode==200) {
+	  var devices=response.body.devices.rows;
+	  //TODO: Issue 63.  Reenable this when devices count is correct
+          //expect(response.body.devices.count).to.equal(expectedDevices.length);
+          expect(devices.length).to.equal(expectedDevices.length);
+	  var dev_count;
+	  var user_count;
+          var sortDevices=sortArrayOn(devices,'devicename');
+          var sortExpectedDevices=sortArrayOn(expectedDevices,'devicename');
+          for (dev_count=0; dev_count < expectedDevices.length; dev_count++) {
```
gferraro: `let devCount=0;` camel casing for variables

integration-tests/cypress/commands/api/device.ts
```diff
@@ -0,0 +1,539 @@
...
+      userName,
+      statusCode
+    ).then((response)=>{
+       if(statusCode==null || statusCode==200) {
+	  var devices=response.body.devices.rows;
+	  //TODO: Issue 63.  Reenable this when devices count is correct
+          //expect(response.body.devices.count).to.equal(expectedDevices.length);
+          expect(devices.length).to.equal(expectedDevices.length);
+	  var dev_count;
+	  var user_count;
```
gferraro: unsued

gferraro: this block also needs to be indented properly

gferraro: where applicabe use let instead of var

integration-tests/cypress/commands/api/device.ts
```diff
@@ -0,0 +1,539 @@
...
+  		  expect(users[count].DeviceUsers.admin).to.equal(expectedUsers[count].DeviceUsers.admin);
+  		  expect(users[count].DeviceUsers.UserId).to.equal(expectedUsers[count].DeviceUsers.UserId);
+	      };
+  	    };
+    	  }
+       };
+  });
+});
+
+Cypress.Commands.add("apiCheckDevices_contains", (userName: string, expectedDevices: [ComparableDevice], params: string = {}, statusCode: number = 200) => {
```
gferraro: `apiCheckDevicesContains`

integration-tests/cypress/commands/api/device.ts
```diff
@@ -0,0 +1,539 @@
...
+  makeAuthorizedRequestWithStatus(
+      {
+        method: "GET",
+        url: fullUrl,
+        body: null
+      },
+      userName,
+      statusCode
+    ).then((response)=>{
+       if(statusCode==null || statusCode==200) {
```
gferraro: indent properly, alsl looks like some very similar parts to above can this be put into smaller chunks of re usable code?

integration-tests/cypress/commands/api/device.ts
```diff
@@ -0,0 +1,539 @@
...
+                };
+              };
+	    });
+            expect(found).to.equal(true);
+          }
+       };
+  });
+});
+
+Cypress.Commands.add("apiCheckDeviceInGroup", (userName: string, cameraName: string, groupName: string, groupId: number, expectedDevice: all, params:string  = {}, statusCode: number = 200) => {
```
gferraro: if gorupId can be null should be spelled out here for tyescript

integration-tests/cypress/commands/api/device.ts
```diff
@@ -0,0 +1,539 @@
...
+        method: "GET",
+        url: fullUrl,
+        body: null
+      },
+      userName,
+      statusCode
+    ).then((response) => {
+       if(statusCode==null || statusCode==200) {
+	  var device=response.body.device;
+	  expect(device.id).to.equal(getCreds(cameraName).id);
```
gferraro, line 245: i'd say these comparison of types, could be put into seperate functions and referenced in all places used

integration-tests/cypress/commands/api/device.ts
```diff
@@ -0,0 +1,539 @@
...
+       type: 'thermalRaw'
+  };
+
+  const fileName="bird_"+id+".cptv";
+  const uniqueName=deviceName+"_"+id;
+  const aliasName=deviceName+"_a"+id;
+
+
+  const creds = getCreds(cameraName);
+  uploadFileRequest(fileName, uniqueName, aliasName, recordingsUrl, data, creds)
```
gferraro: does this work? i would of thought aliasName shoul dbe the deviceName? and it doesn't look like uniqueName is even used. should be deleted?

integration-tests/cypress/commands/api/device.ts
```diff
@@ -0,0 +1,539 @@
...
+Cypress.Commands.add("apiUploadEvent", (cameraName) => {
+  const eventsUrl = v1ApiPath('events');
+  const deviceName = getTestName(cameraName);
+  const creds = getCreds(cameraName);
+
+  const data = {
+       description: {
+	   type: 'throttle'
+       },
+       dateTimes: ['2017-11-13T00:47:51.160Z']
```
gferraro: should this be a nullable parameter and if null set to datetime.now?

integration-tests/cypress/commands/api/device.ts
```diff
@@ -0,0 +1,539 @@
...
+      },
+     cameraName
+    );
+
+});
+
+Cypress.Commands.add("apiCheckDeviceHasRecordings", (username, deviceName,count) => {
+  const user = getCreds(username);
+  const camera = getCreds(deviceName);
+  const fullUrl = v1ApiPath('')+encodeURI('recordings?where={"DeviceId":'+camera.id+'}');
```
gferraro: this would be better written as 
```
const params = {
  where: JSON.stringify({"DeviceId":camera.id})
};

v1ApiPath('recordings`,params)
```

integration-tests/cypress/commands/api/device.ts
```diff
@@ -0,0 +1,539 @@
...
+  }).then((request) => {
+    expect(request.body.count).to.equal(count);
+  });
+});
+
+
+Cypress.Commands.add("apiCheckEventUploaded", (username, deviceName, eventType) => {
+  const user = getCreds(username);
+  const camera = getCreds(deviceName);
+  const eventURL = v1ApiPath('events')+'?deviceId='+camera.id;
```
gferraro: dsimilar to above use the optional query params option for v1ApiPath()

integration-tests/cypress/commands/api/device.ts
```diff
@@ -0,0 +1,539 @@
...
+  cy.request({
+    method: "GET",
+    url: eventURL,
+    headers: user.headers
+  }).then((request) => {
+    expect(request.body.rows[0].EventDetail.type).to.equal(eventType);
```
gferraro: will this fail nicely? or will it cause an exception with no first item of rows existing?

integration-tests/cypress/integration/api/devices/device_in_group.ts
```diff
@@ -0,0 +1,80 @@
+/// <reference path="../../../support/index.d.ts" />
+
```
gferraro, line 2: variables need to be in camelcase as per js style

integration-tests/cypress/integration/api/devices/device_query.ts
```diff
@@ -0,0 +1,168 @@
...
+
+const HTTP_AuthorizationError = 401;
+const HTTP_BadRequest = 400;
+const HTTP_OK = 200;
+
+import { getTestName } from "../../../commands/names";
+import { getCreds } from "../../../commands/server";
+
+describe("Devices/query", () => {
+  const group_admin = "Fredrick-group_admin";
```
gferraro: again camel case variables

integration-tests/cypress/integration/api/devices/device_register.ts
```diff
@@ -0,0 +1,78 @@
+/// <reference path="../../../support/index.d.ts" />
+
+import { getTestName } from "../../../commands/names";
+import { getCreds } from "../../../commands/server";
+
+describe("Device register", () => {
+  const camsGroup = "cams";
+  const otherCams = "other cams";
+
+  const KEEP_DEVICE_NAME = false;
```
gferraro, line 15: why are these constants? and not just set when calling the function

integration-tests/cypress/integration/api/devices/device_reregister.ts
```diff
@@ -0,0 +1,166 @@
...
+const HTTP_BadRequest = 400;
+const HTTP_OK = 200;
+
+import { getTestName } from "../../../commands/names";
+import { getCreds } from "../../../commands/server";
+
+describe("Device reregister", () => {
+  const KEEP_DEVICE_NAME = true;
+  const GENERATE_UNIQUE_NAME = false;
+  const GENERATE_PASSWORD = null;
```
gferraro, line 14: do these need to be ocnstants?

integration-tests/cypress/integration/api/devices/device_users_add_remove.ts
```diff
@@ -0,0 +1,213 @@
...
+  const userC = "Candy";
+  const userD = "Dylan";
+  const camera2 = 'second_H_camera';
+  const superuser = 'admin_test';
+  const su_passwd = 'admin_test';
+  let device_member_details;
+  let device_admin_details;
+  let group_admin_details;
+  let group_member_details;
+  let userB_details;
```
gferraro: camel casing again

gferraro (changes requested):
A few things to be tidied up.
I assumed variables are always camel cased in js but maybe you can confirm with Jon our js style guidelines
Also i think let should always be used instead of var, unless you are using the additional scope of var

integration-tests/cypress/commands/api/device.d.ts
gw7nvw: Corrected

README.md
```diff
@@ -47,7 +47,7 @@ Currently both sets of tests need to be run.
 
 #### Running the cypress tests
 
-See [Cypress Tests](api/test-cypress/README.md) for details on running the cypress tests.
+See [Cypress Tests](integreation-tests/README.md) for details on running the cypress tests.
```
hardiesoft: `integreation` -> `integration`

integration-tests/cypress/commands/api/device.d.ts
hardiesoft: Should `expectedDevice: all` be `expectedDevice: any`? Or a more specific type since it seems like the shape of the device object is known?

integration-tests/cypress/commands/api/device.d.ts
```diff
@@ -34,12 +51,23 @@ declare namespace Cypress {
       makeCameraNameTestName?: boolean
     );
 
+
     /**
     * Retrieve device details using name and groupname
+    * use groupId if provided, otherwise groupName
     * compare with expected device details (JSON equivalent to that retunred by API)
     * optioanlly check for a non-200 status code
```
hardiesoft: `optioanlly` -> `optionally`

integration-tests/cypress/commands/api/device.d.ts
```diff
@@ -34,12 +51,23 @@ declare namespace Cypress {
...
+    apiCheckDeviceInGroup(userName: string, cameraName: string, groupName: string, groupId: number, expectedDevice: all, statusCode: number);
+
+
+    /**
+    * Retrieve devices list
+    * compare with expected device details (JSON equivalent to that retunred by API)
+    * pass optional params (params) to API call
+    * optionlly check for a non-200 status code
+    */
+    apiCheckDevices(userName: string, expectedDevice: [ComparableDevice], params: string, statusCode: number);
```
hardiesoft: ComparableDevice doesn't seem to be defined anymore?

integration-tests/cypress/commands/api/device.d.ts
```diff
@@ -34,12 +51,23 @@ declare namespace Cypress {
...
     */
-    apiCheckDevice(userName: string, cameraName: string, groupName: string, expectedDevice: ComparableDevice, statusCode: number): Chainable<Element>;
+    apiCheckDeviceInGroup(userName: string, cameraName: string, groupName: string, groupId: number, expectedDevice: all, statusCode: number);
+
+
+    /**
+    * Retrieve devices list
+    * compare with expected device details (JSON equivalent to that retunred by API)
+    * pass optional params (params) to API call
+    * optionlly check for a non-200 status code
```
hardiesoft: `optionlly` -> `optionally`

hardiesoft (commented):
Just a few comments, otherwise looks good

integration-tests/cypress/commands/api/device.d.ts
gw7nvw: Yes. Changed.  The 'camera' is old stuff from Clare I've been working on replacing with Device.  Should all be Device now.

integration-tests/cypress/commands/api/device.ts
gw7nvw: Yes.  Changed check for 'not 500 internal server error' to check response matches expected status code

integration-tests/cypress/commands/api/device.ts
gw7nvw: corrected - not sure how that one slipped through lint as it pulled me up on that elsewhere!

integration-tests/cypress/commands/api/device.ts
gw7nvw: yes - good catch

integration-tests/cypress/integration/api/devices/device_reregister.ts
gw7nvw: Was asked to use this approach (rather than just pass true/false) in past code review - to make code 'human readable' without the need to look at the called function to see what the parameter does.  Happy to revert to past behaviour - if that's better. Need 3rd opinion!

integration-tests/cypress/integration/api/devices/device_register.ts
gw7nvw: inherited this approach - its throughout the existing test code so continued for consistency.

integration-tests/cypress/integration/api/devices/device_reregister.ts
hardiesoft: constants are definitely more readable than anonymous true/false args IMO.

integration-tests/cypress/commands/api/device.ts
gw7nvw: No. keepName is used to disable generating a unique name by appending a prefix (so we can test invalid name strings).    Have renamed to doNotMakeNameUnique for clarity

integration-tests/cypress/commands/api/device.ts
gw7nvw: No Existing code saves/looks up names using the base name supplied (which is only used once per test run). Uniqueness is only applied to stop data in the database from multiple test runs clashing.

integration-tests/cypress/commands/api/device.ts
gw7nvw: Agree

integration-tests/cypress/commands/api/device.ts
gw7nvw: yes - pre-existing and used by Clare's code. Will rework to match the standard test structure.

integration-tests/cypress/commands/api/device.ts
gw7nvw: done

integration-tests/cypress/commands/api/device.ts
gw7nvw: done

integration-tests/cypress/commands/api/device.ts
gw7nvw: Yes.  Laziness! Extracted to function & reused.

integration-tests/cypress/commands/api/device.ts
gw7nvw: Please clarify what is required

integration-tests/cypress/commands/api/device.ts
gw7nvw: Noted

integration-tests/cypress/commands/api/device.ts
gw7nvw: Noted.  Part of browser testing code.  Will revisit this when browser testing code is updated. This function and the one used by the API tests in fileUpload.ts do the same job and need to be combined.

integration-tests/cypress/commands/api/device.ts
gw7nvw: Yes.  Noted. This existing function will be rewritten to allow flexible testing when Events tests are written.

integration-tests/cypress/commands/api/device.ts
gw7nvw: Agree

integration-tests/cypress/commands/api/device.ts
gw7nvw: Agree

integration-tests/cypress/commands/api/device.ts
gw7nvw: Agree.  For now have added check to see if rows 0 exists.  But this existing function needs rewriting to allow for multiple events and check the correct values are present in the event. Will be done when event tests are written.

integration-tests/cypress/commands/api/device.ts
gw7nvw: Noted.  But currently there is no consistency of these types between different API endpoints.  So there is no opportunity for reuse.  Standardise the types and we can standardise the testing.

integration-tests/cypress/commands/api/device.ts
gw7nvw: This was a duplicate function from browser tests.  Removed and browser tests pointed at the correct api function..

integration-tests/cypress/commands/api/device.ts
gw7nvw: Removed this duplicate function and refactored tests to use the standard createDevice function and check for correct fail result

integration-tests/cypress/commands/api/device.d.ts
gw7nvw: now defined in types.d.ts and referenced

hardiesoft: Sorry, also is `[ComparableDevice]` meant to be an array of devices?  If so, the correct syntax is `ComparableDevice[]`, and if not, the square brackets should go.

gw7nvw: Yes. Corrected this [] everywhere now

integration-tests/cypress/commands/api/device.ts
gw7nvw: No - nt what keepName was used for. keepName was used by clare to disable unique name generation (so we could test specific strings in the name field).  I have renamed the field to generateUniqueName (and reversed the logic) to avoid future confusion.

integration-tests/cypress/commands/api/device.ts
gw7nvw: Done - optional param in definitions file

integration-tests/cypress/commands/api/device.ts
gw7nvw: types all now declared in types.ts - so code can be reused if we identify any types that are reused

integration-tests/cypress/commands/api/device.ts
gw7nvw: Deleted.  Duplicate function left over from browser test code. Reworked browser tests to use the other version of this function.

integration-tests/cypress/commands/api/device.ts
gw7nvw: Function deleted.  Duplicate function from browser test code.  Browser tests now pointed at the correct version of this function and this function deleted.

PR #16, opened 2021-08-15
Break thumnail loop

hardiesoft:


api/api/V1/recordingUtil.ts
```diff
@@ -183,10 +184,12 @@ async function getCPTVFrame(recording: Recording, frameNumber: number) {
   let finished = false;
   let currentFrame = 0;
   let frame;
-  while (!finished) {
+  const maxFrames = meta.fps * 60 * 10;
```
gferraro: do we need this? or do we specifically not want to work with recordings over 10 minutes

gferraro (approved):
Thanks

PR #17, opened 2021-08-16
Add all events tests

gw7nvw:
Changes for new tests for api/V1/events:
- apidocs corrections and clarifications for api/V1/events
- Tests added for api/V1/events
- Standard format test API wrappers written for all api/V1/events endpoints
- Renamed old non standard-format events test functions & indicated they are deprecated  
- linting fixed throughout integration-tests (mainly around typing)
- browse tests updated to use newly written standard-format api wrappers & old non-standard browse /commands functions removed


api/api/V1/Event.ts
```diff
@@ -92,15 +95,19 @@ export default function (app: Application, baseUrl: string) {
    * @apiGroup Events
    *
    * @apiUse V1UserAuthorizationHeader
-   * @apiParam {Datetime} [startTime] Return only events after this time
+   * @apiParam {Datetime} [startTime] Return only events on after this time
```
hardiesoft: Should this be "on *or* after this time"?

api/api/V1/Event.ts
```diff
@@ -92,15 +95,19 @@ export default function (app: Application, baseUrl: string) {
...
    * @apiParam {Integer} [deviceId] Return only events for this device id
    * @apiParam {Integer} [limit] Limit returned events to this number (default is 100)
    * @apiParam {Integer} [offset] Offset returned events by this amount (default is 0)
    * @apiParam {String} [type] Alphaonly string describing the type of event wanted
    * @apiParam {Boolean} [latest] Set to true to see the most recent events recorded listed first
    *
-   * @apiSuccess {JSON} rows Array containing details of events matching the criteria given.
+   * @apiUse V1ResponseSuccess
+   * @apiSuccess {Number} offset Offset of returned page of results from 1st result matched by query.
+   * @apiSuccess {Number} count Total number of results matchig the query.
```
hardiesoft: `matchig` -> `matching`

integration-tests/cypress/browse/users/group.ts
```diff
@@ -111,14 +111,14 @@ describe("Group Admin Pages", () => {
...
-  function getUserRow(username) {
+  function getUserRow(username: string) {
     // just make sure we are actually targeting the row html first
     cy.get(usersTable).contains(username).parent().should("match", "tr");
     cy.get(usersTable).contains(username).parent().as(username);
     return "@" + username;
   }
 
-  function getGroupPageUrl(groupname) {
+  function getGroupPageUrl(groupname: string) {
```
hardiesoft, line 121: For consistency, `username` should be `userName` and `groupname` should be `groupName` I think.

integration-tests/cypress/browse/users/new_user_and_camera.ts
```diff
@@ -18,15 +18,13 @@ context("Users can see footage from their cameras", () => {
   });
 
   it("A camera event can be triggered", () => {
-    const eventType = "throttle";
     cy.apiSignInAs(username);
-
-    // FIXME Is this supposed to be cy.createExpectedEvent?
-    // cy.apiUploadEvent(camera, "throttle");
-
+    cy.apiEventsAdd(camera, { type: "throttle" });
```
hardiesoft, line 22: Generally I think we should be able to do ```cy.apiEventsAdd(camera, { type: "throttle" }).then((eventDetailId) => {
// Other stuff
});```
Without having the `cy.wait()` call - which would be more error prone.  The cypress linter lists use of arbitrary waits as a warning for this reason.

integration-tests/cypress/commands/api/alerts.d.ts
```diff
@@ -6,7 +6,7 @@ declare namespace Cypress {
     /**
      * Create an alert for a device. Optioanlly expect to fail with code: failCode
      */
-    apiAddAlert(
+    apiAlertAdd(
       user: string,
```
hardiesoft: For consistency, should `user` be `userName`?

integration-tests/cypress/commands/api/events.ts
```diff
@@ -1,49 +1,337 @@
...
 export const EventTypes = {
   POWERED_ON: "rpi-power-on",
   POWERED_OFF: "daytime-power-off",
   STOP_REPORTED: "stop-reported",
 };
 
 Cypress.Commands.add(
-  "recordEvent",
+  "apiEventsAdd",
   (
```
hardiesoft, line 33: The signature of this function is out of sync with its definition in types.d.ts.  This seems to be a common problem, and I'm not asking you to fix this now - it just seems dumb that the way cypress commands are defined it's hard to share a single definition of the function.  I'll do some digging around to see if there's a better way we can do this so that we only have to define function signatures in one place.

integration-tests/cypress/commands/api/monitoring.d.ts
```diff
@@ -14,7 +14,7 @@ declare namespace Cypress {
     checkMonitoring(
       user: string,
       camera: string,
-      expectedVisits: import("../types").TestComparableVisit[],
+      expectedVisits: TestComparableVisit[],
```
hardiesoft, line 17: The reason for this being imported inline was to avoid re-declaration of the type name `TestComparableVisit` in the namespace `Cypress` since this is already declared in `visits.d.ts` I was getting typescript compiler errors.

integration-tests/cypress/integration/api/events/power_events_query.ts
```diff
@@ -0,0 +1,205 @@
+/// <reference path="../../../support/index.d.ts" />
+//
+// This test set verifies correct retrieval of power ewvents
```
hardiesoft: typo `ewvents`

hardiesoft (approved):
Looks very comprehensive, thanks for this!  I'll spend some time once your work on these tests settles down to try and make using typescript with cypress a bit more ergonomic.  I don't think we're really getting the benefits of it here in the same way that we are in the other places we're using it, and I agree we don't want ceremony for the sake of ceremony - i.e. being forced to annotate things as `any` doesn't actually improve anything for anyone.  Approved, and take a look at the few comments I have.  Mostly I'd like to remove `cy.wait` where possible to make the tests a bit less brittle.  At the moment we do get some spurious failures on some runs that seem to be due to timing issues around this.

PR #20, opened 2021-09-06
Add influx metrics

CameronRP:
- Added logging of recordings metrics with influx. Can be viewed at https://monitoring.cacophony.org.nz/d/8d_KP647z/recordings?orgId=1&refresh=30s


docker-compose.yml
```diff
@@ -15,8 +17,5 @@ services:
       - 5432:5432
       - 9001:9001
     volumes:
-      - ./api/models:/app/models
-      - ./api/apidoc:/app/apidoc
-      - ./api/config:/app/config
-      - ./api/api/:/app/api
-      - ./api/scripts:/app/scripts
+      - ./api:/app
+      - node_modules:/app/node_modules
```
hardiesoft, line 21: I think this is likely to break things for people who aren't running on linux - the node_modules shouldn't be mirrored across to the docker image, since locally native modules may be compiled for a different platform and/or architecture.

hardiesoft (changes requested):
Other than the node_modules issue, looks good to me!

hardiesoft (approved):

PR #26, opened 2021-11-07
Api refactor tests

gw7nvw:
Test updated following API refactor and processing changes:
- Testing of split Analyse/Tracking processing stages
- Testing of soft-delete
- Testing of authenticate using deviceId as well as devicename


types/build-schemas.js
```diff
@@ -0,0 +1,233 @@
+"use strict";
```
hardiesoft: This file shouldn't be included - I've added it to .gitignore in my branch so that it won't be picked up in the future.

api/build/config.gypi
```diff
@@ -0,0 +1,83 @@
+# Do not edit. File was generated by node-gyp's "configure" step
```
hardiesoft: This file doesn't look like it should be included in the commit

hardiesoft (changes requested):

hardiesoft (approved):

PR #28, opened 2021-11-10
Api refactor

hardiesoft:


Dockerfile
```diff
@@ -4,12 +4,9 @@
 FROM cacophonyproject/server-base:4
 
 # NOTE: Using locally for arm64
-# FROM cacophony-api
+#FROM hardiesoft/server-base:latest
```
gferraro: guessing this shouldn't be here

gferraro (approved):
Looks all good

PR #29, opened 2021-11-12
Api refactor tests

gw7nvw:
Changes:
- Add tag, track and track-tag tests
- Re-enable checking that all parameters are present when checking data returned by API calls (was disabled before/during API refactor) and rework test data to pass
- Change checks of returned messages[] to check for a substring rather than entire message (check for equal was not practical as message contained ids, etc)
- Save tagJWT against a recording name when retrieving recordings/needs-tag (additional parameter to needs-tag test wrapper)


integration-tests/cypress/commands/types.d.ts
```diff
@@ -315,7 +315,7 @@ export interface ApiRecordingSet {
   metadata?: ApiRecordingDataMetadata;
   additionalMetadata?: ApiThermalAdditionalMetadata | any;
   comment?: string;
-  processingState?: RecordingProcessingState;
+  processingState?: RecordingProcessingState | string;
```
hardiesoft: I think the better thing to do here is actually change the places where it's complaining from strings to the `RecordingProcessingState.XXX` enum variant, otherwise we're just throwing away the type safety.

integration-tests/cypress/commands/types.d.ts
```diff
@@ -286,8 +286,8 @@ export interface ApiRecordingForProcessing {
   fileMimeType: string;
   processingState: string;
   processingMeta: any;
-  GroupId: number;
-  DeviceId: number;
+  GroupId: number | string;
+  DeviceId: number | string;
```
hardiesoft: Why would these ever be (legally) strings instead of numbers?

integration-tests/cypress/integration/api/recordings/recording_processing.ts
```diff
@@ -1224,7 +1323,7 @@ describe("Recordings - processing tests", () => {
 
     it("Recordings in other states not picked up for processing", () => {
       const recording14 = TestCreateRecordingData(templateRecording);
-      recording14.processingState = "FINIHED";
+      recording14.processingState = "FINISHED";
```
hardiesoft: Use the enum variants instead of strings to avoid typos etc.

hardiesoft (changes requested):
Just a few changes around enums vs strings for known string variants, otherwise looks good.  Will be nice to have all this additional coverage!

integration-tests/cypress/commands/types.d.ts
gw7nvw: Yes. But then complains when I test with invalid values.  Just live with the complaint? Or can I define a singleInvalid value as a type and do RecordingProcessingState | invalidValue?

integration-tests/cypress/integration/api/recordings/recording_processing.ts
gw7nvw: done and cast invalid values to accepted type where needed for test

integration-tests/cypress/commands/types.d.ts
gw7nvw: done and cast invalid values to accepted type where needed for test

hardiesoft (approved):

PR #30, opened 2021-11-15
Ability to reset password

gferraro:
Can reset password which will send a url which can be used to change password within 24 hours.
Link will not work if it has already been used to change the password


api/api/V1/User.ts
```diff
@@ -257,4 +258,34 @@ export default function (app: Application, baseUrl: string) {
       euaVersion: config.euaVersion,
     });
   });
+
+  app.patch(
+    `${apiUrl}/changePassword`,
+    validateFields([body("token"), body("password")]),
+    fetchUnauthorizedRequiredUserByResetToken(body("token")),
+    validPasswordOf(body("password")),
```
hardiesoft: This `validPasswordOf(body("password"))` should go where `body("password")` is in the `validateFields` function.

api/api/V1/AuthenticateUser.ts
```diff
@@ -223,4 +224,58 @@ export default function (app: Application) {
...
+          deprecatedField(validNameOf(body("username"))),
+          validNameOf(body("userName")),
+          validNameOf(body("nameOrEmail")),
+          body("nameOrEmail").isEmail(),
+          body("email").isEmail(),
+        ],
+        "Missing user name in request"
+      ),
+    ]),
+    fetchUnauthorizedOptionalUserByNameOrEmailOrId(
```
hardiesoft, line 249: Shouldn't this be `fetchUnauthorizedRequiredUserByNameOrEmailOrId`?  Then no need to check for the user later.

api/api/V1/AuthenticateUser.ts
```diff
@@ -223,4 +224,58 @@ export default function (app: Application) {
       });
     })
   );
+
+  app.post(
```
hardiesoft, line 235: Missing APIdocs

hardiesoft (changes requested):
Looks pretty good, just a couple of API notes to check

api/api/V1/AuthenticateUser.ts
gferraro: This is a security thing so no matter what you never know if you have gotten an invalid or valid user, when you try to reset a password. i.e. always want to return 200. I think this is normal?

PR #32, opened 2021-11-17
add heartbeat endpoint

gferraro:


api/models/Device.ts
```diff
@@ -562,6 +566,13 @@ order by hour;
     return newDevice;
   };
 
+  Device.prototype.updateHeartbeat = async function (nextHeartbeat) {
+    return this.update({
+      lastConnectionTime: new Date(),
```
CameronRP, line 594: I think Jon said that this is automatically updated when a deice uses a JWT?

CameronRP (approved):
Looks good to me.

api/models/Device.ts
hardiesoft: It's automatically updated for the device event-upload and recording upload endpoints, where the device token is used, but not "automatic" in general.

api/api/V1/Device.ts
```diff
@@ -699,4 +699,30 @@ export default function (app: Application, baseUrl: string) {
...
+
+     * @apiUse V1ResponseSuccess
+     * @apiUse V1ResponseError
+     */
+  app.post(
+    `${apiUrl}/heartbeat`,
+    extractJwtAuthorisedDevice,
+    validateFields([body("nextHeartbeat").isISO8601().toDate()]),
+    async function (request: Request, response: Response, next: NextFunction) {
+      response.locals.requestDevice.updateHeartbeat();
```
hardiesoft: You don't appear to be using the user-supplied `request.body.nextHeartbeat` value when updating.

api/models/Device.ts
hardiesoft: (So this update is correct)

api/api/V1/Device.ts
```diff
@@ -699,4 +699,30 @@ export default function (app: Application, baseUrl: string) {
...
+     * @apiUse V1DeviceAuthorizationHeader
+     *
+     * @apiBody {Date} nextHeartbeat time next heart beat is expected
+
+     * @apiUse V1ResponseSuccess
+     * @apiUse V1ResponseError
+     */
+  app.post(
+    `${apiUrl}/heartbeat`,
+    extractJwtAuthorisedDevice,
```
hardiesoft, line 717: This won't actually give you a real `Device` model object, just the id, so you will want to do `const actualDevice = await models.Device.getByPk(response.locals.requestDevice.id)` before you try and call `updateHeartbeat()` on it.

api/migrations/20210929000000-alter-device-columns.js
```diff
@@ -30,11 +30,13 @@ module.exports = {
       queryInterface.removeColumn("Devices", "lastRecordingTime"),
       queryInterface.removeColumn("Groups", "lastRecordingTime"),
       queryInterface.removeColumn("Recordings", "fileSize", Sequelize.INTEGER),
-      queryInterface.removeColumn("Recordings", "rawFileSize", Sequelize.INTEGER),
+      queryInterface.removeColumn(
```
hardiesoft, line 33: Wouldn't you need to start a new migration file for these changes, since these old migrations have already been applied?

hardiesoft (changes requested):
There's a fair bit that needs fixing here

hardiesoft (approved):

PR #40, opened 2022-02-02
Audio list info

Zainrax:


api/migrations/20220110021448-cacophonyIndex-to-Recordings.js
```diff
@@ -0,0 +1,261 @@
...
+                // Create a new track in Track table
+                console.log("Creating Track for", id, begin_s, end_s);
+                const trackId = (
+                  await queryInterface.sequelize.query(
+                    `
+                    INSERT INTO "Tracks" ("RecordingId", "AlgorithmId", "data", "createdAt", "updatedAt")
+                    VALUES (:id, :algorithmId, :data, NOW(), NOW())
+                    RETURNING "id";`,
+                    {
+                      transaction,
```
hardiesoft: I think it might be good to unify the concept of the track "geometry" between audio and thermal recordings.  I see that right now you're using the start and end time values to draw your boxes, but I think it would be good to be doing that from a rectangle structure like the `ApiTrackPosition` interface.  These numbers could just be normalised in the range 0..1, with y being 0 and height being 1 initially - so would draw the same boxes you're drawing now, but when we have frequency information we want to include, it wouldn't require any structural changes.  So you'd have `positions: [{ x: 0.25, width: 0.1, y: 0, height: 1 }]` with just a single "frame" of position info.  Thoughts?  We could also just have a different JSON structure here, and not try to shoehorn it into the same format as the thermal recordings - but it would be good to have the frequency bounds in there from day one.  Finally, do you have any more thoughts on how we can merge calls that are almost certainly from the same bird into a single "track"?  For the morepork tracks that you're creating in this migration, we might want to have a post-processing pass in the nearish future that attempts to merge a bunch of these "call" tracks into a single continuous track?

api/migrations/20220110021448-cacophonyIndex-to-Recordings.js
```diff
@@ -0,0 +1,261 @@
...
+                        }),
+                      },
+                    }
+                  )
+                )[0][0].id;
+                console.log(
+                  "Creating TrackTag for track",
+                  trackId,
+                  species,
+                  liklihood
```
hardiesoft: liklihood -> likelihood

api/migrations/20220110021448-cacophonyIndex-to-Recordings.js
```diff
@@ -0,0 +1,261 @@
...
+          Sequelize.JSONB,
+          { transaction }
+        );
+      }
+
+      // sql get or create new algorithm into DetailSnapshot if does not exist
+      let algorithmId = await queryInterface.sequelize.query(`
+        SELECT "id" from "DetailSnapshots" 
+        WHERE "type"='algorithm' AND "details"->>'algorithm'='sliding_window';`);
+      if (!algorithmId[0].length !== 0) {
```
hardiesoft: This if clause looks super dodgey...

api/migrations/20220110021448-cacophonyIndex-to-Recordings.js
Zainrax: Unifying values across thermal and audio where possible I think is super important, I'm still familiarising myself with the thermal side of things so guidance is always welcome. I simply took the current values and moved them over, but I'll normalise to the duration instead.
I think using normalised values is a much better idea in general, just for frequency we would either have to have to set a field for the max/min frequency at the time of making the track, or determine a static max/min frequency across all future tracks (which currently can't be set with the plugin spectrogram I'm using)
I've thought about merging (purposely put a partial track in the figma design to think about that exact case). There are either automatic, or manual tracks, so I think the simplest solution would be the ability to allow users to delete/change the location of an automatic track by dragging the corners to the right location and simply labelling it as a manual track. The other way  is to have logic around if box1 is close to box2 prompt the user to merge etc. which adds extra steps. So not sure what the nicest user experience is yet.
There will likely need post processing on the morepork inference, and even manual tracks, as consistent data typically builds better models.

api/migrations/20220110021448-cacophonyIndex-to-Recordings.js
Zainrax: agreed

hardiesoft (approved):

PR #41, opened 2022-02-02
Latest users tests

gw7nvw:
Added tests for /users and /admin API endpoints
Added complete set of device heartbeat tests
Added functionality to run shell commands in API environment from tests (e.g. trigger stopped devices check)
Added mailServerStub.ts to api/scripts to allow testing of autogenerated emails


integration-tests/cypress/commands/api/group.d.ts
```diff
@@ -1,6 +1,8 @@
 // load the global Cypress types
 /// <reference types="cypress" />
 
+import { ApiStationResponse } from "../../../../types/api/station";
```
hardiesoft, line 4: Pretty sure this change is going to mess with linting - but not actually break anything either way,

hardiesoft (approved):
Looks good to me, I had one comment, but not that important.

PR #42, opened 2022-03-31
Fix next

gferraro:
Everytime browse is querying for a next and previous recoridng, it is doing a count of all possible query results. This is evidently very slow.

Mkae the previous and next recording queries not count all results


hardiesoft (approved):
LGTM

PR #43, opened 2022-05-15
Thermal pinning

Zainrax:
# Tagging & Spectrogram Improvements
# API Changes
### Change {patch} /api/v1/recordings/:id/tracks/:trackId/:tagId to /api/v1/recordings/:id/tracks/:trackId/tags/:tagId
Undelete endpoint that I created would be treated as a tag id, so made more "RESTful" with tags/:tagId, it also includes updating the "data" through this endpoint's "updates" for additional attributes (gender/maturity)
### Add {patch} /api/v1/recordings/:id/tracks/:trackId/undelete
Removes archived from tracks that were soft-deleted
### Change {delete}  /api/v1/recordings/:id/tracks/:trackId include soft-delete defaults to true
Due to the ability for anyone (with permission to the recording) to delete tracks it's important that it's not permanent if information is still on the track (a.k.a another person's tag). Also makes "undeleting" easier since it's only archived
# Browse Changes
## API
- change removeTrack to deleteTrack
- add undeleteTrack for new endpoint
- add updateTrackTag for adding attributes
## UI
- add zooming, with movable sliders to indicate how much is shown
- spectrogram changes based on the sample rate of the file (this was due to some kiwi files being only 8000 sample rate)
- improved drag time
- Add many more birds and labels
- fixed updating cacophony graph on new recording
- added pinning to tagging (Thermal & audio)
- added sheep tag (thermal)
- added cancel current tag
- mobile friendly playlist
- add map for audio device


hardiesoft (approved):
Let's get this synced up!

PR #45, opened 2022-06-09
Browse UI updates

hardiesoft:
Note that this PR is what is currently deployed to prod - just wanting to get it synced up with main.  There is also a bunch of in-progress stuff there for the browse-next-ui stuff which doesn't yet have tests written for it, but none of that is exercised in current browse, so I think it's fine to ignore that stuff in terms of code review for the moment.


Zainrax (approved):
Looks great! Very exciting to see the browse-next changes

PR #47, opened 2022-06-13
Browse UI fixes

hardiesoft:
- Fixes scheduler picking of devices
- Fixes station links so that you don't get in a state where station reference photos can't be uploaded.
- Adds ability to rename station to station page
- Upgrade/migrate to Cypress 10


Zainrax (approved):
Seems to work just fine. Just about the last PR, everything seems fine, just noticed it creates some untracked files in api/emails, is this intended?

hardiesoft:
> 

Sorry, not sure what you mean with those untracked files - can you give me an example?

PR #48, opened 2022-06-20
Audio improvements

Zainrax:
- Add get /recording/track-tags endpoint to find all track tags for a user/super-sure. Going through query endpoint wasn't an option with the amount of data needed
- Add stats page /monitoring-analysis for all current track tags for Thermal or Audio to give a sense of what models can be built
- Fixed docs, due to one of my previous commit that linted the api docs to shorter than 80 chars per line
- Fixed eslint for typescript as current "extend" ordering caused function types and enum to be checked by eslint rules before typescript rules

Still currently writing tests for endpoint, just getting my head around cypress, but has been tested by Shaun, and been on production without any obvious hitches.


browse/src/api/User.api.ts
```diff
@@ -67,6 +68,33 @@ function list(): Promise<
   return CacophonyApi.get("/api/v1/listUsers");
 }
 
+interface UserData {
+  id: number;
+  userName: string;
+  email: string;
+  firtname?: string;
```
hardiesoft: typo `firtname` -> `firstname`

browse/src/views/MonitoringAnalysisView.vue
```diff
@@ -0,0 +1,567 @@
+<template>
+  <main>
+    <div class="top-bar-container">
+      <h1>Montioring Analysis</h1>
```
hardiesoft: Typo `Montioring` -> `Monitoring`

browse/src/views/MonitoringAnalysisView.vue
```diff
@@ -0,0 +1,567 @@
...
+      ? Array.isArray(station)
+        ? station
+        : [station]
+      : [];
+    devices.value = device ? (Array.isArray(device) ? device : [device]) : [];
+    groups.value = group ? (Array.isArray(group) ? group : [group]) : [];
+    userFocus.value = user ? (user as string) : null;
+
+    onMounted(async () => {
+      await updateTrackTags();
```
hardiesoft, line 409: If this is potentially a slow query, could be good to have some UI state around it to add a spinner etc.

hardiesoft (approved):
A few typos and notes to address, but otherwise looks good, cheers.

PR #50, opened 2022-06-28
Fix user reset

CameronRP:


hardiesoft (approved):
Whoops, sorry for breaking this - I guess our current integration tests don't use this param.  We'll standardise for using email for everything in the near future.

hardiesoft:
Actually, never mind, I see Patrick has already added a fix for this.

PR #51, opened 2022-07-18
fix(Device): visits to recordings path on device link

Zainrax:
Issue with devices page not navigating out of visits to recordings for thermal devices. Only happened on mobile as desktop would navigate to recordings of the device.

Fix is just making the link go straight to recordings instead of visits which I presume are now only associated to stations anyways.

Currently on test, and seems to work.


hardiesoft (approved):
👍

PR #52, opened 2022-07-24
return mass information with track positions

gferraro:


Zainrax (approved):
Looks fine, there isn't anything planned for production anytime soon, so if you want to make a release that would be fine. Let me know if you want me to do that

PR #53, opened 2022-08-14
Changes to allow changing user/display name

hardiesoft:
Plus some refactors to make database column names consistently camelCase for things like userName, deviceName, groupName etc.

Also pulls in a bunch of WIP stuff from my browse-next UI work, but that can be ignored for now, since it doesn't affect other functionality.


Zainrax (approved):
Nice, seems to work well from testing on Thermal and Audio. No issues creating a user, tagging tracks, and then changing the display name.
Non urgent issues:
- "Add user to group" for super admins Modal may need to include emails in case of overlapping display names.

PR #54, opened 2022-09-12
Heirarchical tags

Zainrax:
Recording query's where tags are provided give different results now, and the UI for searching these tags has changed.
Searching by tag will by default include all tags that are under that class of tag (e.g. mustelids will include stoats, and ferrets).  There is an additional "exclusive" query that searches only the tags given same as before.
The structure for the classes are held within two json files, classification.json is a parent to child tree that is for users, and label_paths.json that denotes the paths for a given label when a tag is inserted.

**NOTE**: While testing on browse-test station photos have seemed to disappear after a day, unsure if this is an existing issue, as no files concerning photos has been changed. May be worth backing up the images before releasing if there is nothing obvious causing it.


browse/src/components/ClassificationsDropdown.vue
```diff
@@ -0,0 +1,91 @@
...
+    LayeredDropdown,
+  },
+  setup(props) {
+    const options = ref<Option>({ label: "", children: [] });
+
+    const getClassifications = async () => {
+      const cached = localStorage.getItem("classifications");
+      if (cached) {
+        const parsed = JSON.parse(cached);
+        const res = await api.classifications.getClassifications(
```
hardiesoft: Because you're `await`ing here, you're blocking, and so the cache isn't doing what you intend.  Perhaps it's better to handle the result in `.then`, and return and update the cached option if it changes, but let things go ahead with the cached version until then?

hardiesoft (approved):
Just one comment that would be good to get addressed, though it's not a show-stopping bug.  Thanks!

browse/src/components/ClassificationsDropdown.vue
Zainrax: Agreed and done, that seems like the smarter caching strategy

PR #56, opened 2022-09-19
Heirarchical tags

Zainrax:
Sql "Exists" is a lot slower than "limit 1 is not null". We only need to know if a single tag matches for a given recording query.

### DIFF
--- a/api/models/Recording.ts
+++ b/api/models/Recording.ts
@@ -858,9 +858,9 @@ from (
               Sequelize.fn(
                 "json_build_object",
                 "start_s",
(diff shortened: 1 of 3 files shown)


api/models/Recording.ts
```diff
@@ -849,18 +849,18 @@ from (
         {
           model: models.Track,
           where: trackWhere,
-          required: trackRequired,
-          separate: true,
+          required: false,
+          seperate: true,
```
hardiesoft: typo `seperate` -> `separate`

hardiesoft (approved):
Just a typo to fix

PR #58, opened 2022-10-10
New thumbs

gferraro:
Add thumbnail per track
Choose best thumbnail based of track tags using Visit logic at point of query
Add optional trackid to thumbnail endpoint


hardiesoft (approved):
LGTM, let's get it on test and see if there are any issues

PR #61, opened 2023-01-12
Add list of devices to ignore when making a service error report

CameronRP:
This add the ability to have a list of devices where there errors won't be reported on the daily email.


api/scripts/report-service-errors.ts
```diff
@@ -16,10 +16,24 @@ async function main() {
     endTime: endDate.toDate().toISOString(),
     startTime: startDate.toDate().toISOString(),
   };
-  const serviceErrors = await errors(
+  var serviceErrors = await errors(
```
hardiesoft: Typically we use `let` instead of `var`.  It may not matter in this context, but it can lead to unintended variable scoping issues.

api/scripts/report-service-errors.ts
```diff
@@ -16,10 +16,24 @@ async function main() {
...
     { query: query, res: { locals: { requestUser: {} } } },
     true
   );
+  const ignoredDevices = config.deviceErrorIgnoreList;
+  if (ignoredDevices) {
+    var prunedServiceErrors = {} as ServiceErrorMap;
+    for (const [key, serviceError] of Object.entries(serviceErrors)) {
+      for (const device of serviceError.devices) {
+        if (ignoredDevices.indexOf(device) == -1) {
```
hardiesoft: Might be clearer as `if (!ignoredDevices.includes(device)) { ... }`

hardiesoft (approved):
Mostly looks good, thanks.  Will be good to have less noise in those daily emails!

PR #63, opened 2023-01-25
Feat/delete account

Zainrax:


api/api/V1/User.ts
```diff
@@ -666,6 +667,35 @@ export default function (app: Application, baseUrl: string) {
     }
   );
 
+  app.delete(
+    `${apiUrl}/request-delete-user`,
+    extractJwtAuthorizedUser,
```
hardiesoft, line 672: Because you're logging in with a long (infinite) lived JWT token, even once the user is deleted it's possible that they can hit various API endpoints with their token, and have things maybe sorta kinda work (Assuming the endpoint doesn't try to lookup their user.)  While not a requirement for this feature, ideally we want to deprecate the old authorize endpoint that hands out these tokens (https://api.cacophony.org.nz/#api-Authentication-AuthenticateUser), and have clients use the `/api/v1/users/authenticate-user` end-point - which hands out tokens with a 5 minute lifespan, plus a refresh token which can be swapped for an up to date JWT token, more in line with best practices for using JWTs for authentication.  Not saying you need to do this now for the side-kick app, since it's a bit more work, but it's something you should be thinking about.

api/emails/transactionalEmails.ts
```diff
@@ -358,6 +358,39 @@ export const sendGroupMembershipRequestEmail = async (
...
+    adminEmail.text,
+    "coredev@cacophony.org.nz",
+    `User ${userEmailAddress} has requested deletion of their account`
+  );
+
+  return await sendEmail(
+    userEmail.html,
+    userEmail.text,
+    userEmailAddress,
+    "Your Cacophony Monitoring account has been deleted",
```
hardiesoft: Maybe change the subject of the email to "Your request to delete your Cacophony Monitoring account has been received"?

hardiesoft (approved):

PR #66, opened 2023-07-12
Linear freq

gferraro:
Adjust the frequencies to be correct and tracks to be created in linear scale, rather than log scale with origin at (0,0)

### DIFF
--- a/api/scripts/migrate-audio.ts
+++ b/api/scripts/migrate-audio.ts
@@ -1,0 +1,97 @@
+import process from "process";
+import { program } from "commander";
+import pkg from 'pg';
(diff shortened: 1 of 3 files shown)


Zainrax (approved):

api/scripts/migrate-audio.ts
```diff
@@ -0,0 +1,109 @@
+import process from "process";
+import { program } from "commander";
+import pkg from "pg";
+const { Client } = pkg;
```
Zainrax, line 4: Bit of an odd way to import. `import { Client } from "pg";` not possible? not that important though

PR #69, opened 2023-08-15
Add human filter, audio page improvements

Zainrax:
Changes:
- Add filterHumans to group settings json
- Add redacted to Recording model
- Add increased volume for audio
- No longer replay audio when selecting or modifying an audio track
- Fix allow groups without devices


api/config.cjs
hardiesoft, line 1: Does this need tidying up?

hardiesoft (approved):
This all looks great to me!

api/config.cjs
Zainrax: Yep, thanks for pointing that out

PR #70, opened 2023-09-05
Add ir player

gferraro:


browse/src/components/Video/ThermalVideoPlayer.vue
```diff
@@ -0,0 +1,585 @@
...
+      if (currentTrackOnly) {
+        if (this.tracks.length && this.tracks[this.currentTrack.trackIndex]) {
+          tracks = [this.tracks[this.currentTrack.trackIndex]];
+        } else {
+          tracks = [];
+        }
+      } else {
+        tracks = this.tracks;
+      }
+      const frameTime = 1 / 10;
```
hardiesoft: Seems like assumptions about framerate wouldn't be true for non-cptv video?

hardiesoft (approved):
Just a couple of comments - would like to understand what re-track means in terms of CPTV files – I thought it was just the on-device ir tracking that needed cleanup?  Is the on-device CPTV tracking also bad?

Otherwise looks good to me, thanks!

browse/src/components/Video/ThermalVideoPlayer.vue
gferraro: No but didnt think we'd eve use this for cptv files, i'll add a check for the type anyway

hardiesoft (approved):

PR #72, opened 2023-09-25
Audio improvements

Zainrax:
API changes:
- stations return recording count
- added comment to recording tag
Browse:
- Adds Notes section that is a Tag with a comment (No tag just defaults to 'note")
- Ability to delete stations


api/api/V1/Station.ts
```diff
@@ -152,6 +153,41 @@ export default function (app: Application, baseUrl: string) {
...
+    `${apiUrl}/:id/recordings-count`,
+    extractJwtAuthorizedUser,
+    validateFields([
+      idOf(param("id")),
+      query("view-mode").optional().equals("user"),
+    ]),
+    async (request: Request, response: Response) => {
+      const stationdId = request.params.id;
+
+      response.locals.recordingsCount = await models.Recording.count({
```
hardiesoft: Not quite sure why you're stashing this in response.locals, as opposed to a locally scoped variable?

api/api/V1/Station.ts
```diff
@@ -152,6 +153,41 @@ export default function (app: Application, baseUrl: string) {
...
+   */
+  app.get(
+    `${apiUrl}/:id/recordings-count`,
+    extractJwtAuthorizedUser,
+    validateFields([
+      idOf(param("id")),
+      query("view-mode").optional().equals("user"),
+    ]),
+    async (request: Request, response: Response) => {
+      const stationdId = request.params.id;
```
hardiesoft: Strictly speaking should probably be making sure the user has access to the station – not that we're going to leak much data here, but still...

api/log.log
```diff
@@ -0,0 +1,3 @@
+[32minfo[39m: Sending email with subject Stopped Devices to cy_jerry_c82032a0@api.created.com
+[31merror[39m: Error: connection encountered an error (connect ECONNREFUSED 127.0.0.1:7777)
```
hardiesoft: Guessing you didn't mean to commit this file?

hardiesoft (approved):
A few comments to address, but otherwise cool.

PR #74, opened 2023-10-09
Monitoring Stats

Zainrax:
## Api Changes
- **get** /api/v1/recordings/track-tags/count
- **post** /api/v1/stations/recordings-count

## Features
- Add track tag statistics to users at group level ("Stats" tab in group), and across all in stats at the top (/monitoring-analysis)
- Add a composite index on deletedAt and duration, and archivedAt on track tag since it's always used. Duration check does remove corrupted recordings, but I've changed to a is not null check instead.

## Fixes
- Remove access for emails on super admin which have "read" only, includes removing the ability to view as different user.
- Changes to tab list to hide using a custom ":show" instead of using v-if, which was causing routing issues.
- Display all "master" track tags in audio track list, previously only showed first

### DIFF
--- a/api/api/V1/Recording.ts
+++ b/api/api/V1/Recording.ts
@@ -1198,7 +1198,6 @@ export default (app: Application, baseUrl: string) => {
       csv.writeToStream(response, rows);
     }
   );
-  
 
   /**
    * @api {get} /api/v1/recordings/:id Get a recording
@@ -1280,7 +1279,6 @@ export default (app: Application, baseUrl: string) => {
       }
     }
   );
-
 
   /**
    * @api {get} /api/v1/recordings/track-tags/count Get track tag counts
--- a/api/api/V1/Recording.ts
+++ b/api/api/V1/Recording.ts
@@ -1198,6 +1198,90 @@ export default (app: Application, baseUrl: string) => {
       csv.writeToStream(response, rows);
     }
   );
+  
+
+  /**
+   * @api {get} /api/v1/recordings/:id Get a recording
+   * @apiName GetRecording
+   * @apiGroup Recordings
+   *
+   * @apiUse MetaDataAndJWT
+   * @apiUse V1UserAuthorizationHeader
+   *
+   * @apiUse V1ResponseSuccess
+   *
+   * @apiParam {Integer} id Id of the recording to get.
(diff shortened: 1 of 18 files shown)


api/api/V1/Recording.ts
```diff
@@ -1279,6 +1280,75 @@ export default (app: Application, baseUrl: string) => {
...
+      query("exclude").default([]).optional().isArray(),
+      query("includeAI").default(false).isBoolean(),
+      integerOf(query("offset")).optional(),
+      integerOf(query("limit").optional()),
+      query("type")
+        .default("thermalRaw")
+        .optional()
+        .isIn(Object.values(RecordingType)),
+      query("view-mode").optional().equals("user"),
+      integerOf(query("groupId")).optional(), // Added validation for groupId
```
hardiesoft, line 1329: A style note:  at a glance at this API I can't tell that you're validating that the request user belongs to the group (in the non-super-user case).  I dug through the getTrackTagsCount and I see that it does in fact add the correct constraint to the SQL that it generates – but that's an internal implementation detail.  Generally speaking I think it's good to make those contracts obvious as part of the API validation boilerplate i.e. using `fetchAuthorizedRequiredGroupById` – *even* though that's at the expense of a tiny amount of efficiency.  

As some background, in the distant past we had validation kind of threaded through the entire api/model ad-hoc, and it was extremely difficult to work out by looking at an API end-point whether validation existed, and if so, where in the call-stack.  It also took a long time to untangle, and I'd like to see us not go back there!

api/api/V1/Station.ts
```diff
@@ -152,6 +153,68 @@ export default function (app: Application, baseUrl: string) {
...
+   * @apiBody {String} [from-date] Date the station should be active from
+   */
+  app.post(
+    `${apiUrl}/recordings-count`,
+    extractJwtAuthorizedUser,
+    validateFields([
+      body("stationIds")
+        .isArray({ min: 1 })
+        .withMessage("At least one station ID is required"),
+      query("view-mode").optional().equals("user"),
```
hardiesoft, line 178: We don't seem to use view-mode at all in the following logic, so I guess this end-point doesn't really work in super-user mode to view anyone's station recording counts?

hardiesoft (approved):
Looks good, just some notes around api validation style.

PR #76, opened 2023-11-13
Megadetector

gferraro:
Add support for trail cam analysis.
And draw tracks in browse


Zainrax (approved):
What is the additional "animal" tag is for? Does mammal not cover it, or is this due to some AI tagging that's going on?

Zainrax (approved):

PR #78, opened 2023-11-26
Rat thresh

gferraro:
Added processing endpoint to get rat threshold data at a point in time.

Updated device history endpoints to use device id in query

Added script to generate rat threshold data


hardiesoft (approved):
Looks good to me

PR #81, opened 2023-12-12
Masked regions API

willgeorgeson:


api/api/V1/Device.ts
```diff
@@ -1062,6 +1063,120 @@ export default function (app: Application, baseUrl: string) {
     }
   );
 
+  /**
+ * @api {post} /api/v1/devices/:deviceId/mask-regions Set mask regions for a device
+ * @apiName SetDeviceMaskRegions
+ * @apiGroup Device
+ * @apiParam {Integer} deviceId Id of the device
+ * @apiBody {Object[]} maskRegions collection of mask regions
```
hardiesoft: Usually when we have a complex, known structure to be passed to the API - we validate that what we receive actually conforms to the structure/type, and in terms of the automatically generated API docs, we output what the type is: See around line 452: `@apiInterface {apiSuccess::ApiDeviceResponseSuccess} device Device details`, and where it is defined around line 173: 
```
// eslint-disable-next-line @typescript-eslint/no-unused-vars
interface ApiDeviceResponseSuccess {
  device: ApiDeviceResponse;
}
```

We have a plugin that parses this and outputs the actual type in the docs: https://api.cacophony.org.nz/#api-Device-GetDeviceById

api/api/V1/Device.ts
```diff
@@ -1062,6 +1063,120 @@ export default function (app: Application, baseUrl: string) {
...
+        });
+
+        return successResponse(response, "Mask regions added successfully");
+      } else {
+        return successResponse(
+          response,
+          "No device history settings entry found to add mask regions"
+        );
+      }
+    } catch (e) {
```
hardiesoft: Either remove this try {} catch {} block, or make sure that a response is returned in the error case – otherwise the API call will hang.

api/api/V1/Device.ts
```diff
@@ -1062,6 +1063,120 @@ export default function (app: Application, baseUrl: string) {
...
+          where: {
+            fromDateTime: deviceHistoryEntry.fromDateTime,
+            uuid: device.uuid,
+            GroupId: device.GroupId,
+          }
+        });
+
+        return successResponse(response, "Mask regions added successfully");
+      } else {
+        return successResponse(
```
hardiesoft, line 1217: Probably we want some kind of error response in the instance there is not a device history entry to update.

api/api/V1/Device.ts
```diff
@@ -1062,6 +1063,120 @@ export default function (app: Application, baseUrl: string) {
...
+ * @apiSuccess {String} message Success message
+ * @apiUse V1ResponseError
+ */
+
+app.post(
+  `${apiUrl}/:id/mask-regions`,
+  extractJwtAuthorizedUser,
+  validateFields([
+    idOf(param("id")),
+    body("maskRegions").isArray().not().isEmpty(),
```
hardiesoft: For complex known structures, we'd typically validate them here, rather than the validation be just "is this an array".

See line ~1206 of this file
```
body("setStationAtTime").custom(
        jsonSchemaOf(ApiDeviceLocationFixupSchema)
      ),
``` 
for an example of how the JSON schema validation looks.  Update the expected type of `DeviceHistorySettings` to include the sub-type you've defined for the masked regions payload.

api/api/V1/Device.ts
```diff
@@ -1062,6 +1063,120 @@ export default function (app: Application, baseUrl: string) {
...
+ * @apiName GetDeviceSettings
+ * @apiGroup Device
+ * @apiParam {Integer} deviceId Id of the device
+ *
+ * @apiDescription Retrieves settings for a device from the DeviceHistory table.
+ *
+ * @apiUse V1UserAuthorizationHeader
+ *
+ * @apiUse V1ResponseSuccess
+ * @apiSuccess {Object} settings Device settings
```
hardiesoft: Include the structure of the returned object in the docs.

api/api/V1/Device.ts
```diff
@@ -1062,6 +1063,120 @@ export default function (app: Application, baseUrl: string) {
...
+          uuid: device.uuid,
+          GroupId: device.GroupId,
+          location: { [Op.ne]: null },
+          fromDateTime: { [Op.lte]: atTime },
+        },
+        order: [["fromDateTime", "DESC"]],
+      });
+
+      if (deviceSettings) {
+        return successResponse(response, "Device mask-regions retrieved successfully", deviceSettings.settings);
```
hardiesoft: I'd expect that we'd want to return just the mask-regions part of the device settings, not the entire device settings object.

api/api/V1/Device.ts
```diff
@@ -1062,6 +1063,120 @@ export default function (app: Application, baseUrl: string) {
...
+      } else {
+        return successResponse(response, "No device mask-regions found");
+      }
+    } catch (e) {
+      console.log(e);
```
hardiesoft: Not sure the try catch is needed – what can fail?  If it is needed, we need to return a response in the failure case.

api/models/DeviceHistory.ts
```diff
@@ -45,10 +45,15 @@ export interface DeviceHistorySettings {
     bottomRight: [number, number];
   };
   maskPolygons?: {
```
hardiesoft: I think you called this `maskRegions` when you saved it in your post request, not `maskPolygons`.
Also, no need anymore for optional `exclude` and `label` fields I think?  Also, I think the points member you're saving is actually `points: { x: number, y: number }[],` right?

browse-next/src/api/Device.ts
```diff
@@ -380,6 +380,21 @@ export const getReferenceImageForDeviceAtCurrentLocation = (
...
+  const params = new URLSearchParams();
+  params.append("at-time", atTime.toISOString());
+  return CacophonyApi.get(
+    `/api/v1/devices/${deviceId}/mask-regions?${params}`
+  ) as Promise<FetchResult<any>>;
+};
+
+export const updateMaskRegionsForDevice = (
+  deviceId: DeviceId,
+  maskRegionsData: { maskRegions: { region: string; points: { x: number; y: number }[] }[] },
```
hardiesoft: Worth naming this type and pulling it into the shared front-end/back-end types at `/types/api/device.d.ts/`

browse-next/src/api/Device.ts
```diff
@@ -380,6 +380,21 @@ export const getReferenceImageForDeviceAtCurrentLocation = (
   ) as Promise<FetchResult<any>>;
 };
 
+export const getMaskRegionsForDevice = (deviceId: DeviceId, atTime: Date) => {
+  const params = new URLSearchParams();
+  params.append("at-time", atTime.toISOString());
+  return CacophonyApi.get(
+    `/api/v1/devices/${deviceId}/mask-regions?${params}`
+  ) as Promise<FetchResult<any>>;
```
hardiesoft: We know the return type, so use it instead of `any`.

browse-next/src/assets/camera-view.jpeg
hardiesoft, line 1: I'm not sure we want to commit this?

browse-next/src/components/DeviceSetupDefineMask.vue
```diff
@@ -0,0 +1,823 @@
...
+    null
+  );
+});
+
+const getExistingMaskRegions = async () => {
+  if (device.value) {
+    const existingMaskRegions = await getMaskRegionsForDevice(device.value.id);
+    console.log("Lenbth", existingMaskRegions.result.maskRegions.length);
+    for (let i = 0; i < existingMaskRegions.result.maskRegions.length; i++) {
+      points.value = existingMaskRegions.result.maskRegions[i].points;
```
hardiesoft: It's not clear why you're putting points in a ref value and then sticking that into another ref?

browse-next/src/components/DeviceSetupDefineMask.vue
```diff
@@ -0,0 +1,823 @@
...
+
+function formatRegionData(regionsArray) {
+  const jsonStructure = {
+    "maskRegions": []
+  };
+
+  for (let i = 0; i < regionsArray.value.length; i++) {
+    const extractedPoints = regionsArray.value[i].regionData;
+    const regionObj = {
+      "region": i.toString(),
```
hardiesoft: I think we should allow user-named regions, rather than just using a number.

browse-next/src/components/DeviceSetupDefineMask.vue
```diff
@@ -0,0 +1,823 @@
...
+      points.value = existingMaskRegions.result.maskRegions[i].points;
+      regionsArray.value.push({ regionData: points.value });
+    }
+    console.log("regionsArray: ", regionsArray);
+  }
+};
+
+const updateExistingMaskRegions = async () => {
+  const data = JSON.parse(formatRegionData(regionsArray));
+  const outcome = await updateMaskRegionsForDevice(device.value.id, data);
```
hardiesoft: This probably wants to handle a possible error response.

browse-next/src/components/DeviceSetupDefineMask.vue
```diff
@@ -0,0 +1,823 @@
...
+  return regionsArray.value.length > 0;
+});
+
+const computeImageDimensions = () => {
+  isMobile.value = window.innerWidth < mobileWidthThreshold;
+  const img = document.querySelector(".imageContainer");
+  if (img) {
+    const viewportWidth = window.innerWidth;
+    if (!isMobile.value) {
+      cptvFrameWidth.value = 160 * (viewportWidth / 450);
```
hardiesoft, line 120: This still feels quite brittle – I think `cptvFrameWidth` and `cptvFrameHeight` seem to only depend on `viewportWidth`, so it would be better to have a `vueUse` handler on the window width, i.e. resize, dpiChange, and have the cptvFrame dimensions be computed properties that automatically update when that fires.

Same with the `canvasElement` width and height - these can be computed properties that then get updated in the template when changed, rather than imperatively.

browse-next/src/components/DeviceSetupDefineMask.vue
```diff
@@ -0,0 +1,823 @@
...
+    const clickedX = event.clientX - left;
+    const clickedY = event.clientY - top;
+    const values = normalise(clickedX, clickedY);
+    points.value.push({ x: values.x, y: values.y });
+    drawPolygon();
+  }
+}
+
+function drawRegionIndices() {
+  const canvasElement = canvas.value;
```
hardiesoft: Because the canvas ref is typed as `Ref<HTMLCanvasElement | null>` you'll want to cast this to `HTMLCanvasElement` since you're sure it exists (assuming you only call this function *after* `onMounted`), so that the linter doesn't complain.

browse-next/src/components/DeviceSetupDefineMask.vue
```diff
@@ -0,0 +1,823 @@
...
+    if (i === 0) {
+      context.moveTo(x, y);
+    } else {
+      context.lineTo(x, y);
+    }
+  }
+  context.stroke();
+
+  context.fillStyle = "rgba(13, 110, 253, 1)";
+  for (let i = 0; i < points.value.length; i++) {
```
hardiesoft: In this kind of instance where you don't actually need the index of the point `for (const point of points.value) { ... }` is a more readable form.

browse-next/src/components/DeviceSetupDefineMask.vue
```diff
@@ -0,0 +1,823 @@
...
+        firstPoint.y * canvas.value.height - lastPoint.y * canvas.value.height,
+        2
+      )
+  );
+
+  if (distance < polygonClosedTolerance.value) {
+    points.value[points.value.length - 1] = points.value[0];
+  }
+
+  clearMask(); //causes them to snap as desired, however recursion error
```
hardiesoft: I wouldn't expect to see these side-effects in what should be a pure computed function.

browse-next/src/components/DeviceSetupDefineMask.vue
```diff
@@ -0,0 +1,823 @@
...
+      2
+    ) +
+      Math.pow(
+        firstPoint.y * canvas.value.height - lastPoint.y * canvas.value.height,
+        2
+      )
+  );
+
+  if (distance < polygonClosedTolerance.value) {
+    points.value[points.value.length - 1] = points.value[0];
```
hardiesoft, line 336: You probably don't want to be mutating other values inside this computed function.  Do this elsewhere once you've decided if the polygon is closed or not.

browse-next/src/components/DeviceSetupDefineMask.vue
```diff
@@ -0,0 +1,823 @@
...
+  updateExistingMaskRegions();
+  selectingArea.value = !selectingArea.value;
+  points.value = [];
+  clearMask();
+  generateMask();
+};
+
+const toggleMaskEnabled = () => {
+  maskEnabled.value = !maskEnabled.value;
+  if (maskEnabled.value) {
```
hardiesoft: This might be better as watcher on `maskEnabled`?

browse-next/src/components/DeviceSetupDefineMask.vue
```diff
@@ -0,0 +1,823 @@
...
+
+const toggleInclusionRegion = () => {
+  inclusionRegion.value = !inclusionRegion.value;
+  generateMask();
+};
+
+function removePoint(): void {
+  if (points.value.length > 1) {
+    points.value.pop();
+    drawPolygon();
```
hardiesoft: Not sure you need to `drawPolygon` here when you're just going to clear the mask and redraw it anyway.

browse-next/src/components/DeviceSetupDefineMask.vue
```diff
@@ -0,0 +1,823 @@
...
+  points.value = [];
+  clearMask();
+  generateMask();
+};
+
+const toggleMaskEnabled = () => {
+  maskEnabled.value = !maskEnabled.value;
+  if (maskEnabled.value) {
+    points.value = [];
+    clearMask();
```
hardiesoft: `clearMask` gets called in either case, so maybe move it outside the conditional?

browse-next/src/components/DeviceSetupDefineMask.vue
```diff
@@ -0,0 +1,823 @@
...
+  clearMask();
+  generateMask();
+  drawPolygon();
+}
+
+function addRegionSelection(): void {
+  regionsArray.value.push({ regionData: points.value });
+  points.value = [];
+  toggleCreatingRegion();
+  const canvasElement = canvas.value;
```
hardiesoft: Looks like you have `clearMask` for this.

browse-next/src/components/DeviceSetupDefineMask.vue
```diff
@@ -0,0 +1,823 @@
...
+  polygonClosedTolerance.value = devicePixelRatio.pixelRatio.value * 10;
+  if (regionsArray.value.length === 0) {
+    return;
+  }
+  const maskCanvas = canvas.value;
+  const maskContext = maskCanvas.getContext("2d");
+  maskCanvas.width = cptvFrameWidth.value * devicePixelRatio.pixelRatio.value;
+  maskCanvas.height = cptvFrameHeight.value * devicePixelRatio.pixelRatio.value;
+
+  if (inclusionRegion.value) {
```
hardiesoft: Does inclusion/exclusion data get saved to the back-end at the moment?

browse-next/src/components/DeviceSetupDefineMask.vue
```diff
@@ -0,0 +1,823 @@
...
+    regionsArray.value.forEach((region) => {
```
hardiesoft: It seems like it might have been easier to keep the region data in the same format that you're saving to the backend, rather than needing a step where you convert between the two?

browse-next/src/components/DeviceSetupDefineMask.vue
```diff
@@ -0,0 +1,823 @@
...
+    });
+  } else {
+    maskContext.beginPath();
+    maskContext.moveTo(0, 0);
+    maskContext.lineTo(maskCanvas.width, 0);
+    maskContext.lineTo(maskCanvas.width, maskCanvas.height);
+    maskContext.lineTo(0, maskCanvas.height);
+    maskContext.lineTo(0, 0);
+
+    regionsArray.value.forEach((region) => {
```
hardiesoft: You may want to destructure `region` when you iterate over the array: `regionsArray.value.forEach(({regionData}) => {`

browse-next/src/components/DeviceSetupDefineMask.vue
```diff
@@ -0,0 +1,823 @@
...
+
+function deleteRegion(index: number) {
+  regionsArray.value.splice(index, 1);
+  clearMask();
+  generateMask();
+}
+
+function toggleCreatingRegion(): void {
+  creatingRegion.value = !creatingRegion.value;
+  if (creatingRegionAborted.value) {
```
hardiesoft, line 465: `creatingRegionAborted` probably didn't need to be a reactive `ref` - it looks like it should just be an argument to the `toggleCreateRegion` function.

browse-next/src/components/DeviceSetupDefineMask.vue
```diff
@@ -0,0 +1,823 @@
...
+          </b-button>
+        </div>
+      </div>
+    </div>
+  </div>
+</template>
+
+<style scoped>
+@media screen and (max-width: 767px) {
+  .contentContainer {
```
hardiesoft: By convention, CSS class names should be in kebab-case.

integration-tests/cypress/e2e/api/devices/devices.ts
```diff
@@ -1,8 +1,35 @@
 /// <reference path="../../../support/index.d.ts" />
 import { getTestName } from "@commands/names";
 import { makeAuthorizedRequest, v1ApiPath, getCreds } from "@commands/server";
+import { TestGetLocation } from "@commands/api/station";
```
hardiesoft: Roll back your changes to this file now that you've broken things out into a separate file.

hardiesoft (changes requested):
I've done a first pass and given a bit of feedback, yell out if anything is unclear.

browse-next/src/components/DeviceSetupDefineMask.vue
willgeorgeson: I'm not quite sure what you mean here, which line specifically?

browse-next/src/components/DeviceSetupDefineMask.vue
willgeorgeson: Does this instance not depend on the index? I'm not sure how you'd achieve the same behaviour without including the index

browse-next/src/components/DeviceSetupDefineMask.vue
willgeorgeson: It doesn't yet, I'm thinking we have a conversation about this because there's a few different ways you could do it. Also would be good to talk about labelling regions too.

api/api/V1/Device.ts
```diff
@@ -165,6 +167,11 @@ interface ApiDeviceLocationFixupBody {
   setStationAtTime: ApiDeviceLocationFixup;
 }
 
+// // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
hardiesoft: You'll want to uncomment this.

api/api/V1/Device.ts
```diff
@@ -1070,6 +1077,7 @@ export default function (app: Application, baseUrl: string) {
  * @apiParam {Integer} deviceId Id of the device
  * @apiBody {Object[]} maskRegions collection of mask regions
```
hardiesoft: Remove redundant @apiBody and @apiBodyExample lines.

api/api/V1/Device.ts
```diff
@@ -1070,6 +1077,7 @@ export default function (app: Application, baseUrl: string) {
  * @apiParam {Integer} deviceId Id of the device
  * @apiBody {Object[]} maskRegions collection of mask regions 
  * @apiBodyExample {json} 
+ * @apiInterface {apiBody::MaskRegionsDataBody} recording The recording data.
```
hardiesoft: The comment on this `recording The recording data.` seems copy-pasted from somewhere, and doesn't make sense here

api/api/V1/Device.ts
```diff
@@ -1100,40 +1110,142 @@ app.post(
...
-      } else {
-        return successResponse(
-          response,
-          "No device history settings entry found to add mask regions"
-        );
-      }
+      return successResponse(response, "Mask regions added successfully");
     } catch (e) {
       console.log(e);
+      return next(
```
hardiesoft: Usually we don't use `console.log` for things that are designed to be permanent output on the production server.

We're using winston for structured logging, so prefer something like `logger.error("Error: %s", e);` here.

api/api/V1/Device.ts
```diff
@@ -1100,40 +1110,142 @@ app.post(
...
+        )
+      );
     }
   }
 );
 
+/**
+ * @api {get} /api/v1/devices/:deviceId/mask-regions Get device mask-regions 
+ * @apiName GetDeviceMaskRegions
+ * @apiGroup Device
```
hardiesoft: Would be good to include documentation for the optional `at-time` parameter.

api/api/V1/Device.ts
```diff
@@ -1100,40 +1110,142 @@ app.post(
...
+ *         { "x": 0.8, "y": 0.4 }
+ *       ]
+ *     }
+ *   ]
+ * }
+ *
+ * @apiUse V1UserAuthorizationHeader
+ *
+ * @apiUse V1ResponseSuccess
+ * @apiSuccess {apiSuccess::MaskRegionsData}
```
hardiesoft: This should be `@apiInterface` rather than `@apiSuccess`.

api/api/V1/Device.ts
```diff
@@ -1167,16 +1278,15 @@ app.get(
       });
 
       if (deviceSettings) {
-        return successResponse(response, "Device mask-regions retrieved successfully", deviceSettings.settings);
+        return successResponse(response, "Device settings retrieved successfully", deviceSettings.settings);
```
hardiesoft: Was this API just added to facilitate testing, or does it have another usage?

browse-next/src/components/DeviceSetupDefineMask.vue
```diff
@@ -60,21 +60,28 @@ const device = computed<ApiDeviceResponse | null>(() => {
...
-    const existingMaskRegions = await getMaskRegionsForDevice(device.value.id);
-    console.log("Lenbth", existingMaskRegions.result.maskRegions.length);
-    for (let i = 0; i < existingMaskRegions.result.maskRegions.length; i++) {
-      points.value = existingMaskRegions.result.maskRegions[i].points;
+    const existingMaskRegions = await getMaskRegionsForDevice(
+      device.value.id,
+      mostRecentTime
+    );
+    // calculate number of regions present
+    while (existingMaskRegions.result[regionsLength]) {
```
hardiesoft: Why do we need to calculate this?  Can't we just use the array length?

browse-next/src/components/DeviceSetupDefineMask.vue
```diff
@@ -60,21 +60,28 @@ const device = computed<ApiDeviceResponse | null>(() => {
...
+      regionsLength++;
+    }
+    for (let i = 0; i < regionsLength; i++) {
+      points.value = existingMaskRegions.result[i].points;
```
hardiesoft: Still unclear to me why we want to nest reactive values like this?  Why assign to points.value just to copy it to regionData in the next line, then points.value gets overwritten by the next iteration?

browse-next/src/components/DeviceSetupDefineMask.vue
```diff
@@ -84,8 +91,8 @@ function formatRegionData(regionsArray) {
   for (let i = 0; i < regionsArray.value.length; i++) {
     const extractedPoints = regionsArray.value[i].regionData;
     const regionObj = {
-      "region": i.toString(),
-      "points": extractedPoints
+      region: i.toString(),
```
hardiesoft: Can we please implement user defined region names?  It may also make more sense for the stored regionData to be in the format { "name": pointsArray1, "name2": pointsArray2 }, so that we don't have to iterate through the array of regions to find the one we want to edit by name? 

I also still think that you may be over-complicating the code by having two separate storage formats one on the front-end and another on the back-end, and having to have code to convert between them.

api/api/V1/Device.ts
```diff
@@ -1100,40 +1110,142 @@ app.post(
...
+
+
 /**
  * @api {get} /api/v1/devices/:deviceId/settings Get device settings
  * @apiName GetDeviceSettings
  * @apiGroup Device
  * @apiParam {Integer} deviceId Id of the device
  *
- * @apiDescription Retrieves settings for a device from the DeviceHistory table.
+ * @apiDescription Retrieves mask regions for a device from the DeviceHistory table.
```
hardiesoft: Description seems incorrect.

api/api/V1/Device.ts
```diff
@@ -1100,40 +1110,142 @@ app.post(
...
+      where: {
+        uuid: device.uuid,
+        GroupId: device.GroupId,
+        location: { [Op.ne]: null },
+        fromDateTime: { [Op.lte]: atTime },
+      },
+      order: [["fromDateTime", "DESC"]],
+    });
+
+    if (deviceSettings) {
```
hardiesoft: We need to check for the existence of the `maskRegions` property within settings.

api/api/V1/Device.ts
```diff
@@ -1100,40 +1110,142 @@ app.post(
...
+      return successResponse(response, "Device mask-regions retrieved successfully", deviceSettings.settings.maskRegions);
+    } else {
+      return successResponse(response, "No device mask-regions found");
```
hardiesoft: Still not sure this should be success as opposed to an error?  If success, then include the maskRegions return value as an empty array.

api/api/V1/Device.ts
```diff
@@ -1100,40 +1110,142 @@ app.post(
...
+        uuid: device.uuid,
+        GroupId: device.GroupId,
+        location: { [Op.ne]: null },
+        fromDateTime: { [Op.lte]: atTime },
+      },
+      order: [["fromDateTime", "DESC"]],
+    });
+
+    if (deviceSettings) {
+      return successResponse(response, "Device mask-regions retrieved successfully", deviceSettings.settings.maskRegions);
```
hardiesoft: We'd usually name the returned object, so it would look more like:
```
return successResponse(response, "Device mask-regions retrieved successfully", { "maskRegions": deviceSettings.settings.maskRegions });
      ```
I think.

hardiesoft (changes requested):
A few things still to tidy up, let me know if you need more clarity on anything, cheers!

Would also be good to investigate why Travis is failing on this branch and get that fixed.

api/api/V1/Device.ts
willgeorgeson: I thought there would be instances where we may want to retrieve all the settings within deviceHistory, not just maskRegions. I have used it in my testing to prove that the mask-regions API allows the rest of the settings to persist.

browse-next/src/components/DeviceSetupDefineMask.vue
willgeorgeson: I tried that initially, however would always be undefined. I wasn't sure why it was undefined, but tried for a while and couldn't get it to work so have used that solution. If you can get it to work I would prefer to use .length

hardiesoft: What is the actual structure of the object returned by the api function `getMaskRegionsForDevice`?  It's typed as `Promise<FetchResult<MaskRegionsData>>` which means you should first check that `success` is true, to know what the payload type should be.

PR #85, opened 2024-02-06
Email notifications review

willgeorgeson:


api/api/V1/User.ts
```diff
@@ -356,34 +320,38 @@ export default function (app: Application, baseUrl: string) {
...
-      await requestUser.update({
-        settings: {
-          ...currentSettings,
-          onboardTracking: updatedOnboardTracking,
-        },
-      });
+app.patch(
+  `${apiUrl}/user-settings`,
+  extractJwtAuthorizedUser,
+  validateFields([body("settings").isObject()]),
```
hardiesoft: I'm not clear on why we have these new APIs, I though we were just going to re-use the existing https://api.cacophony.org.nz/#api-User-UpdateUser API?  Note that this api `api/api/V1/User.ts (@api {patch} /api/v1/users)` also already does validation of the user settings object passed, rather than allowing an arbitrary object.

browse-next/src/api/User.ts
```diff
@@ -142,26 +142,32 @@ export const updateUserFields = (
     FetchResult<void>
   >;
 
-export const updateUserOnboarding = (fields: { settings?: ApiUserSettings }) =>
-  CacophonyApi.patch("/api/v1/users/onboarding", fields) as Promise<
+export const updateUserSettings = (fields: { settings?: ApiUserSettings }) =>
```
hardiesoft, line 145: Can't we use the existing `browse-next/src/models/LoggedInUser.ts:112 setLoggedInData` functionality, which in turn calls `browse-next/src/api/User.ts:134 saveUserSettings`?

browse-next/src/views/DashboardView.vue
```diff
@@ -39,10 +39,10 @@ import { canonicalLatLngForLocations } from "@/helpers/Location";
 import { sortTagPrecedence } from "@models/visitsUtils";
 import type { StationId as LocationId } from "@typedefs/api/common";
 import {
-  setUserOnboarding,
-  updateUserOnboarding,
-  getUserOnboarding,
+  updateUserSettings,
+  getUserSettings,
 } from "@/api/User";
+import { cptvplayercontext_streamComplete } from "@/components/cptv-player/cptv-decoder/decoder/decoder_bg.wasm";
```
hardiesoft: This import doesn't seem like it is supposed to be here?

types/api/user.d.ts
```diff
@@ -19,21 +19,30 @@ export interface ApiUserSettings {
...
+    savedSearchQueries?: string[];
+    displayMode?: "audio" | "thermal";
+    viewAsSuperUser?: boolean;
+    currentSelectedGroup?: {
+      groupName: string;
+      id: GroupId;
     };
-    propertyValues: {
-      type: "boolean";
+    onboardTracking?: {
```
hardiesoft: This is non-nonsensical and doesn't do what you thought it does:
Correct syntax would be something like:
```
onboardTracking?: Record<string, boolean>,
emailNotifications?: Record<string, boolean>
```

api/scripts/daily-digest.ts
```diff
@@ -69,34 +62,40 @@ import { Op } from "sequelize";
...
+      const speciesListWidth = `calc(100% / ${speciesListArray.length})`;
+      const speciesListStyle = `border: 2px solid #666666; width: 100%; margin-top: 10px; overflow: hidden; text-align: center; font-size: 0;`;
+      
       const emailData = {
         text: text,
         from: "Cacophony <>",
         to: `${user.userName} <${user.email}>`,
         subject: "Daily digest",
-        attachment: [{ data: html, alternative: true }],
+        attachment: [{ data: html.replace('<div id="speciesListContainer"', `<div id="speciesListContainer"`), alternative: true }],
```
hardiesoft: What's up with this string replacement - replacing a string with the exact same string?

api/scripts/daily-digest.ts
```diff
@@ -38,14 +31,14 @@ import { Op } from "sequelize";
         'id': group.GroupId
```
hardiesoft: Looks like you're making the incorrect assumption that there's only one group/project per user.  I think we'd want to be breaking down the digest for *all* of the users projects individually, but in a single email.

hardiesoft (changes requested):
I've provided enough feedback for you to get on with for a while, but I'm going to have to run this locally and go through it with a finer-toothed comb to properly evaluate all the changes.

api/api/V1/User.ts
willgeorgeson: This is my mistake, I misinterpreted the behaviour of api/api/V1/User.ts (@api {patch} /api/v1/users) thinking it required all the details in the validateFields() function. I have just realised it only required one of them. I'll update it to utilise this API endpoint. Sorry about that.

api/scripts/daily-digest.ts
```diff
@@ -0,0 +1,125 @@
...
+
+(async () => {
+  try {
+    const models = await modelsInit();
+    const users = await models.User.findAll({where: {
+      'settings.emailNotifications.dailyDigest': true
+    }});
+    
+    const templateFilename = "daily-digest.html";
+    const client = new SMTPClient(config.smtpDetails);
```
hardiesoft: This probably wants to be using the `sendEmail` function, which also handles attaching any images, and handling any send errors.

api/scripts/daily-digest.ts
```diff
@@ -0,0 +1,125 @@
...
+
+    users.forEach(async user => {
+      const requestUser = user;
+      const groups = await models.GroupUsers.findAll({where: {
+        'UserId': user.id
+      }});
+
+      groups.forEach(async group => {
+        const recordingData = {};
+        const name = await models.Group.findOne({where: {
```
hardiesoft: You've already got the group object, so you shouldn't need to query the DB again here just to get its name – you already have that on the `group` variable.

api/scripts/daily-digest.ts
```diff
@@ -0,0 +1,125 @@
...
+        
+        const emailData = {
+          text: text,
+          from: config.smtpDetails.from_name,
+          to: `${user.userName} <>`,
+          subject: "Daily digest",
+          attachment: [{ data: html.replace('<div id="speciesListContainer"', `<div id="speciesListContainer"`), alternative: true }],
+        };
+
+        client.send(emailData, (err, message) => {
```
hardiesoft: You seem to be sending a separate digest email for each group a user is a member of.  I think it would be better to have the digest email be a single email breaks down the activity per group.

api/scripts/daily-digest.ts
```diff
@@ -0,0 +1,125 @@
...
+          templateFilename,
+          interpolants
+        );
+        
+        const emailData = {
+          text: text,
+          from: config.smtpDetails.from_name,
+          to: `${user.userName} <>`,
+          subject: "Daily digest",
+          attachment: [{ data: html.replace('<div id="speciesListContainer"', `<div id="speciesListContainer"`), alternative: true }],
```
hardiesoft: Same comment as before – what's up with the html string replace, which just replaces the search string with the same string?

api/scripts/daily-digest.ts
```diff
@@ -0,0 +1,125 @@
...
+
+        const { text, html } = await createEmailWithTemplate(
+          templateFilename,
+          interpolants
+        );
+        
+        const emailData = {
+          text: text,
+          from: config.smtpDetails.from_name,
+          to: `${user.userName} <>`,
```
hardiesoft: Seems like we're setting the user name here, but not the actual email we want to send to!

api/scripts/daily-digest.ts
```diff
@@ -0,0 +1,125 @@
...
+        const visitsTotal = (visits as unknown as Array<any>).length;
+        const speciesListArray = Object.entries(recordingData).map(([species, count]) => {
+          return { species, count, widthPercent: 100 / Object.keys(recordingData).length };
+        });
+
+        const interpolants: {
+          [key: string]: any;
+        } = {
+          groupName: `${name.groupName}`,
+          groupURL: `https://browse-next.cacophony.org.nz/${name.groupName}`,
```
hardiesoft: Group/Project names can be anything, and they're not necessarily URL safe:  There's a function `api/emails/htmlEmailUtils.ts:urlNormaliseName` to deal with this.

api/scripts/daily-digest.ts
```diff
@@ -0,0 +1,125 @@
...
+          'GroupId': group.GroupId
+        }});
+        const stationsArray = [];
+        stations.forEach(async station => {
+          stationsArray.push(station.id);
+        });
+
+        const oneWeekAgo = new Date();
+        const now = new Date();
+        oneWeekAgo.setHours(oneWeekAgo.getHours() - 168);
```
hardiesoft: Seems like this might be clearer using `setDays`?

api/scripts/daily-digest.ts
```diff
@@ -0,0 +1,125 @@
...
+        const recordingData = {};
+        const name = await models.Group.findOne({where: {
+          'id': group.GroupId
+        }});
+
+        const stations = await models.Station.findAll({where: {
+          'GroupId': group.GroupId
+        }});
+        const stationsArray = [];
+        stations.forEach(async station => {
```
hardiesoft: Why is this function `async`?  The's nothing async going on in the function body.

api/scripts/daily-digest.ts
```diff
@@ -0,0 +1,125 @@
...
+      groups.forEach(async group => {
+        const recordingData = {};
+        const name = await models.Group.findOne({where: {
+          'id': group.GroupId
+        }});
+
+        const stations = await models.Station.findAll({where: {
+          'GroupId': group.GroupId
+        }});
+        const stationsArray = [];
```
hardiesoft: This might be more idiomatic as `const stationsArray = stations.map(station => station.id)`;

api/scripts/daily-digest.ts
```diff
@@ -0,0 +1,125 @@
...
+        searchDetails.compareAi = "Master";
+        searchDetails.types = params.types;
+      
+        const visits = await generateVisits(
+          requestUser.id,
+          searchDetails,
+          viewAsSuperAdmin
+        );
+        
+        (visits as unknown as Array<any>).forEach(async visit => {
```
hardiesoft: Once again, no reason for the closure here to be `async`

hardiesoft: I'd like to see the actual typing here, as the return value of `generateVisits` does have a well known, well defined type – which also has an error variant that should be handled.  Typing it as unknown is asking for bugs to be introduced.

api/scripts/weekly-digest.ts
```diff
@@ -0,0 +1,125 @@
...
+    const templateFilename = "weekly-digest.html";
+    const client = new SMTPClient(config.smtpDetails);
+    // const client = new SMTPClient({
+    //   user: "",
+    //   password: "",
+    //   host: "smtp.gmail.com",
+    //   ssl: true,
+    // });
+
+    users.forEach(async user => {
```
hardiesoft: Given that weekly and daily digests share the exact same logic, and only differ by their timespan, maybe pull the shared logic out into a helper function?

api/scripts/weekly-digest.ts
```diff
@@ -0,0 +1,125 @@
...
+
+        client.send(emailData, (err, message) => {
+          console.log(err || message);
+        });
+      });
+
+    });
+
+  } catch (error) {
+    console.error("An error occurred:", error);
```
hardiesoft: We need to use the structured logging logger we use on the backend for logging, not `console.log`

hardiesoft (changes requested):
Some more feedback around the digest emails – I think we're getting there!

api/scripts/daily-digest.ts
willgeorgeson: The Groups table contains the groupName but does not have the associated users in the group. The GroupUsers contains the associated users but not the groupName. What I'm doing is finding all the groups that the specific user is part of from the GroupUsers table, then from the groups the user is a part of I am getting the the group ID and using that to find the name of the group. Since I need both the group name and the group users I think both queries are necessary as neither table contains both pieces of information.

hardiesoft: Ah, right, so what you'd usually do is do that in one query, using the include param i.e: 
```
include: [
      {
        model: models.Group,
        attributes: ["id", "groupName"],
        required: true,
}],
```

api/scripts/daily-digest.ts
willgeorgeson: That goes within the chevrons, however I just removed it when I commited as it was my personal email

PR #90, opened 2024-02-14
- Fix email text for stopped devices report when the stopped device i…

hardiesoft:
…s a bird monitor.

Fixes issue reported by Matthew


willgeorgeson (approved):
Looks good to me

PR #91, opened 2024-04-16
Device updates

Zainrax:
Includes current production changes by @hardiesoft

Additional changes primarily in [this commit](https://github.com/TheCacophonyProject/cacophony-web/commit/94b204d82eeded5f9fb9701b107b8e2f511af88f):
- New /reregister-authorized post to allow users to reassign a device back to a group with a new password (must have the users token, and the device token)
- Added extractJwtAuthorizedUserFromBody to get a authorized token from request body
- Change password hasing for devices from validation to beforeModify so that it is rehashed on modification
- allow /:id/settings get with authorised device token


hardiesoft (approved):
I'll approve this now, but I'll probably re-work the jwt extraction middleware in the not-too-distant future so that we can more easily share that logic rather than having to copy/paste it just because the source field is different.  Looks good, thanks!  Seems like this should make device management much more logical for users!

PR #92, opened 2024-04-16
False trigger

gferraro:
Set audio tracks made in browse with classify flag,
Allow getoneforprocessing to look for recordings with tracks that require classification


hardiesoft (approved):
Just revert the changes to `app_TEMPLATE.js` and then looks good to me

PR #94, opened 2024-05-01
Browse next upgrades

hardiesoft:


api/api/V1/Device.ts
```diff
@@ -1393,7 +1393,7 @@ export default function (app: Application, baseUrl: string) {
 
   app.get(
     `${apiUrl}/:id/settings`,
-    extractJwtAuthorizedUserOrDevice,
+    extractJwtAuthorizedUser,
```
Zainrax: We probably want to allow devices to use this endpoint as well, though it may not be setup properly for devices due to fetchAuthorizedRequiredDeviceById being present(?).

Zainrax (approved):

PR #97, opened 2024-08-01
Feat sync settings, Update latest release to master

Zainrax:


hardiesoft:
@Zainrax Is this still a draft?

api/models/DeviceHistory.ts
```diff
@@ -113,5 +121,93 @@ export default function (
...
+function mergeSettings(
+  currentSettings: ApiDeviceHistorySettings,
+  incomingSettings: ApiDeviceHistorySettings,
+  setBy: DeviceHistorySetBy
+): { settings: ApiDeviceHistorySettings; changed: boolean } {
+  const mergedSettings: ApiDeviceHistorySettings = { ...currentSettings };
+
+  let changed = false;
+  for (const key in incomingSettings) {
+    if (incomingSettings.hasOwnProperty(key)) {
```
hardiesoft, line 252: Small nit: when is this ever not true, given that we're iterating keys in `incomingSettings`.  In the case where you're worried about inherited keys from up the prototype chain you can use `Object.keys` which doesn't iterate inherited keys, only ones that are `hasOwnProperty`.

types/api/device.d.ts
```diff
@@ -51,10 +51,25 @@ export interface ApiDeviceLocationFixup {
...
+};
+
+export type ThermalRecordingSettings = {
+  useLowPowerMode: boolean;
+} & SettingsBase;
+
+export type WindowsSettings = {
+  startRecording: string;
+  stopRecording: string;
+  powerOn: string;
```
hardiesoft: `powerOn` and `powerOff` fields are ignored in DOC AI cameras, so not sure if we should include these at all.  I don't know that we'll ever go back to using them.

types/api/device.d.ts
```diff
@@ -51,10 +51,25 @@ export interface ApiDeviceLocationFixup {
...
+  powerOff: string;
+} & SettingsBase;
+
 export interface ApiDeviceHistorySettings {
+  location?: LatLng;
```
hardiesoft: I assume this would only be different from the `DeviceHistory.location` field in the `synced === false` case?

browse-next/src/views/DeviceDiagnosticsSubView.vue
```diff
@@ -46,11 +63,137 @@ const deviceConfig = ref<LoadedResource<DeviceConfigDetail>>(null);
...
+    if (response.success) {
+      return response.result.settings;
+    }
+  } catch (e) {
+    console.error(e);
+  }
+  return null;
+};
+
+const handleToggleUseLowPowerMode = async (setting: { name: string }) => {
```
hardiesoft: With all of these toggles, there needs to be some kind of "change in progress" variable that disables the button while the change is happening, otherwise users will be able to mash the buttons while the request is in flight.

browse-next/src/views/DeviceDiagnosticsSubView.vue
```diff
@@ -46,11 +63,137 @@ const deviceConfig = ref<LoadedResource<DeviceConfigDetail>>(null);
...
+  }
+};
+
+const enableCustomRecordingWindows = () => {
+  if (!settings.value || !settings.value.windows) {
+    return;
+  }
+  const windows = settings.value.windows;
+  customSettings.value = {
+    powerOn: windows.powerOn,
```
hardiesoft: As noted elsewhere, I think we should probably remove `powerOn/Off`

browse-next/src/views/DeviceDiagnosticsSubView.vue
```diff
@@ -454,7 +767,123 @@ const deviceLocationPoints = computed<NamedPoint[]>(() => {
...
+          :items="settingsTable"
+          :fields="fields"
+          striped
+          hover
+          class="settings-table mt-3"
+        >
+          <template #cell(actions)="row">
+            <div v-if="row.item.name === 'Use Low Power Mode'">
+              <b-button
+                @click="handleToggleUseLowPowerMode(row.item)"
```
hardiesoft: Toggling low power mode should only be available on devices that support it.  One way to work out what kind of device we're looking at currently is to check the `saltNodeGroup` variable on this page for a substring of `tc2`.

browse-next/src/views/DeviceDiagnosticsSubView.vue
```diff
@@ -454,7 +767,123 @@ const deviceLocationPoints = computed<NamedPoint[]>(() => {
       </div>
     </div>
     <div class="mt-4" v-if="device.type === 'thermal'">
+      <h6 class="mt-4">Device Settings:</h6>
```
hardiesoft: This really needs a different view for mobile responsive breakpoints – however, I'm also inclined to want to move this all to the Setup tab, since it seems more appropriate there, so I can take care of that if you want.

browse-next/src/views/DeviceDiagnosticsSubView.vue
```diff
@@ -454,7 +767,123 @@ const deviceLocationPoints = computed<NamedPoint[]>(() => {
...
+              >
+              <b-button
+                @click="handleSet24HourRecordingWindows"
+                :variant="
+                  currentWindowsType === '24hour' ? 'primary' : 'secondary'
+                "
+                >24 Hours</b-button
+              >
+              <b-button
+                @click="enableCustomRecordingWindows"
```
hardiesoft: In my testing, `enableCustomRecordingWindows` did nothing, I assume it's because there were not already window settings defined in my device settings, so it returned early?  That case probably wants to be handled.

hardiesoft (changes requested):
Hey Patrick, thanks for this.  Looks mostly pretty good, but I did find a few things that didn't work quite right, so let me know when you've addressed those issues and I can re-review :)

hardiesoft (approved):

PR #99, opened 2024-10-21
retry failed recordings once after 12 hours

gferraro:
These generally happen with API is down so no point retrying straight away


hardiesoft (approved):
Any chance you can merge these changes with the `device-events-listing` branch?  That's pretty far ahead of `main` at this point, but nobody has reviewed it for me so we can update `main` to it.

Edit: That's now been merged into main.

PR #102, opened 2024-12-10
Retry failed recordings

gferraro:
Rather than using reprocess endpoint which was mean't for changing tracking algorithms and always reprocessing from scratch. Added new endpoint retry failed, for retry a recordings current failed state.
Changed browse to use this endpoint.

### DIFF
--- a/api/api/V1/Reprocess.ts
+++ b/api/api/V1/Reprocess.ts
@@ -39,7 +39,7 @@ export default (app: Application, baseUrl: string) => {
   const apiUrl = `${baseUrl}/reprocess`;
 
   /**
-   * @api {get} /api/v1/reprocess/retry/:id Retry processing a single recording which is in a failed state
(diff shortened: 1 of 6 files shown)


api/api/V1/Reprocess.ts
```diff
@@ -38,12 +38,50 @@ const { uniq: dedupe } = lodash;
...
+   * @apiParam {Integer} id of recording to retry
+   * @apiDescription Retries processing a recording thats in a failed state
+   *
+   * @apiUse V1UserAuthorizationHeader
+   *
+   * @apiUse V1ResponseSuccess
+   * @apiUse V1ResponseError
+   */
+  app.get(
+    `${apiUrl}/retryFailed/:id`,
```
hardiesoft: For url endpoints, can we stick to `kebab-case` for aesthetic and consistency reasons?

hardiesoft (changes requested):
Just a style change, otherwise looks good to me.

hardiesoft (approved):

PR #105, opened 2025-01-29
only update is filtered once processing is finished

gferraro:


hardiesoft (approved):
This looks like it does what you intended, approving :)

hardiesoft:
Actually, I just noticed that `integration-tests/cypress/e2e/api/batch-2/recordings/recording_processing.ts` has most of the tests disabled (something with `it.only` in there) – and if they're enabled a bunch break due to these changes.  Can you please fix the tests before merging this?

PR #108, opened 2025-04-15
Add state

gferraro:
Added new state TrackAndAnalyse as the default state of uploaded recordings.
Added a new processing endpoint ${apiUrl}/:id/tracks/:trackId/thumbnailInfo for updating thumbnail regions specifically when tracking is done on the PI
Added an additionalMetadata field to indicate the source othe track metadata only wil be set from pi at the moment as "PI"
Adjusted logic for if metadata tracking has been supplied


hardiesoft (approved):
Looks good, feel free to merge into my branch, but please don't deploy it to prod as there is some WIP stuff from me in there that's not ready yet.

PR #112, opened 2025-06-04
Request group access & allow non-admin users update station photos and name

Zainrax:
Added new apis to streamline access to groups:
1. Added  users/request-device-access to access group by device id. The device id is only being used to get the current group.
2. modified request-group-membership to make admin email optional so it defaults to owner
These may need ability down the line for admin/owner to disable requests. At the moment non-verified users can't request access to a group.

Allow non-admin users to modify station name, and upload reference photo.
@api {patch} /api/v1/stations/:id/name Update station name
This is mainly due to the original api being a catch all so modifying that didn't seem safe, adding a single endpoint for name keeps things narrow.


hardiesoft (approved):
This all looks solid to me.  One thing to be aware of (though it doesn't affect what you've done) is that there can be multiple project owners.  That's fine, since your code will just pick an arbitrary owner, but in the case where an owner doesn't get the request email as expected, that'd probably be the reason why.

PR #121, opened 2025-12-15
Bulk track tags

gferraro:


api/api/V1/Processing.ts
```diff
@@ -352,10 +367,193 @@ export default function (app: Application, baseUrl: string) {
...
+   * (#FileProcessing:Algorithm) request
+   *
+   * @apiUse V1ResponseSuccess
+   * @apiSuccess {int[]} trackIds of the newly created track.
+   *
+   * @apiUse V1ResponseError
+   *
+   */
+  app.post(
+    `${apiUrl}/:id/tracksAndTags`,
```
hardiesoft: Can we keep api endpoints with the consistent kebab-case style that's generally seen as being more standard for urls?  `tracksAndTags` -> `tracks-and-tags`

api/api/V1/Processing.ts
```diff
@@ -352,10 +367,193 @@ export default function (app: Application, baseUrl: string) {
...
+        );
+      }
+      const data = response.locals.data;
+      const trackIds: number[] = [];
+      const tracks = [];
+      for (const trackData of data) {
+        const isMasked = await isTrackMasked(
+          recording,
+          trackData,
+          request.body.algorithmId,
```
hardiesoft: `isTrackMasked` doesn't seem to use the algorithmId at all, so this arg should be removed.

api/api/V1/Processing.ts
```diff
@@ -352,10 +367,193 @@ export default function (app: Application, baseUrl: string) {
...
+    trackData: MinimalTrackRequestData,
+    algorithmId: number,
+  ): Promise<boolean> => {
+    const deviceId = recording.DeviceId;
+    const groupId = recording.GroupId;
+    const atTime = recording.recordingDateTime;
+    if (recording.type === RecordingType.ThermalRaw) {
+      const positions = trackData && trackData.positions;
+      if (positions) {
+        return trackIsMasked(deviceId, groupId, atTime, positions);
```
hardiesoft, line 497: `trackIsMasked` is an async function because it needs to lookup the `DeviceHistory` to see if there was masking.  If we're adding lots of tracks for a single recording/device at once, we're duplicating this effort unnecessarily.  It would be better to refactor this to pull out the mask region once, and then test against it for every track.

api/api/V1/Processing.ts
```diff
@@ -352,10 +367,193 @@ export default function (app: Application, baseUrl: string) {
...
+  };
+
+  const prepareTrackToSave = (
+    recording: Recording,
+    trackData: MinimalTrackRequestData,
+    algorithmId: number,
+  ): MinimalTrack => {
+    const newTrack = {
+      AlgorithmId: algorithmId,
+      startSeconds: trackData.start_s || 0,
```
hardiesoft: In reality, can we ever have a situation where processing is not setting `start_s` and `end_s`?  I think in the type we have those fields optional for historical reasons, since it wasn't always true that these were always set, but for new data we probably have stronger guarantees right?  Should we maybe make those fields non-optional?

api/api/V1/Processing.ts
```diff
@@ -352,10 +367,193 @@ export default function (app: Application, baseUrl: string) {
...
+            UserId: null,
+            used,
+          } as TrackTag;
+          trackTags.push(tag);
+
+          trackTagData.push(predData);
+        }
+
+        delete trackData.predictions;
+        await Track.saveTrackData(track.id, trackData);
```
hardiesoft: It's probably a bit faster to collect all these promises into an array and then `await` them with `Promise.all` after the loop, so they're saved in parallel as much as possible.

api/api/V1/Processing.ts
```diff
@@ -352,10 +367,193 @@ export default function (app: Application, baseUrl: string) {
...
+      }
+
+      const modelTrackTags = await TrackTag.bulkCreate(trackTags);
+      for (let i = 0; i < modelTrackTags.length; i++) {
+        const modelTrackTag = modelTrackTags[i];
+        if (modelTrackTag.model) {
+          // Save the additional Track metadata to object storage
+          await Track.saveTrackTagData(modelTrackTag.id, trackTagData[i]);
```
hardiesoft: Again, it's probably a bit faster to collect all these promises into an array and then`await` them with `Promise.all` after the loop, so they're saved in parallel as much as possible.

api/api/V1/Processing.ts
```diff
@@ -352,10 +367,193 @@ export default function (app: Application, baseUrl: string) {
...
+      newTrack.maxFreqHz = trackData.maxFreq || 0;
+    }
+    const track = await recording.addTrack(newTrack);
+    return track.id;
+  };
   /**
    * @api {post} /api/v1/processing/:id/tracks Add track to recording
    * @apiName PostTrack
    * @apiGroup Processing
+   * @apiDeprecated Use /api/v1/processing/:id/tracksAndTags
```
hardiesoft, line 556: Given that Processing is the only consumer of this API, and we control that, do we even need to deprecate this?  Maybe just remove entirely now?

api/api/V1/Processing.ts
```diff
@@ -467,6 +635,8 @@ export default function (app: Application, baseUrl: string) {
    * @api {post} /api/v1/processing/:id/tracks/:trackId/tags Add tag to track
    * @apiName PostTrackTag
    * @apiGroup Processing
+   * @apiDeprecated Use /api/v1/processing/:id/tracks/:trackId/tagsBulk
```
hardiesoft: Again, do we need to deprecate, or can we just remove?

api/api/V1/Processing.ts
```diff
@@ -521,6 +691,93 @@ export default function (app: Application, baseUrl: string) {
...
+   *
+   * @apiParam {JSON} data Describing track tags.
+   *
+   * @apiUse V1ResponseSuccess
+   * @apiSuccess {int[]} trackTagIds Unique ids of the newly created track tags.
+   *
+   * @apiUse V1ResponseError
+   */
+  app.post(
+    `${apiUrl}/:id/tracks/:trackId/tagsBulk`,
```
hardiesoft: kebab-case please

api/api/V1/Processing.ts
```diff
@@ -521,6 +691,93 @@ export default function (app: Application, baseUrl: string) {
     },
   );
 
+  /**
+   * @api {post} /api/v1/processing/:id/tracks/:trackId/tagsBulk Add tags to track
```
hardiesoft: What's the use case for tagsBulk as well as tracksAndTags?  Is this mostly for re-processing?

api/api/V1/Processing.ts
```diff
@@ -521,6 +691,93 @@ export default function (app: Application, baseUrl: string) {
...
+        }
+
+        const trackTagIds = [];
+        const modelTrackTags = await TrackTag.bulkCreate(trackTags);
+        for (let i = 0; i < modelTrackTags.length; i++) {
+          const modelTrackTag = modelTrackTags[i];
+          trackTagIds.push(modelTrackTag.id);
+          if (modelTrackTag.model) {
+            // Save the additional Track metadata to object storage
+            await Track.saveTrackTagData(modelTrackTag.id, trackTagData[i]);
```
hardiesoft: Same comment as above about awaiting all the promises at once, rather than sequentially.

api/config/app_TEMPLATE.js
```diff
@@ -93,6 +93,5 @@ export default {
   cacophonyUserIds,
   cacophonyGroupIds,
   processingUserIds,
-  groupNamesWithRedactedThermalRecordings,
```
hardiesoft, line 96: Why was this removed?  Was it intentional?

types/api/fileProcessing.d.ts
```diff
@@ -47,22 +47,25 @@ export interface TrackFramePosition {
 }
 
 interface TrackClassification {
-  classify_time: Seconds;
-  label: ClassificationClass;
+  classify_time?: Seconds;
+  label: string;
   confidence: FloatZeroToOne;
```
hardiesoft: I guess `confidence` should just be typed as a number now?

types/api/fileProcessing.d.ts
```diff
@@ -90,6 +93,65 @@ export interface ThumbnailInfo {
   median_diff: number;
   score: number;
 }
+
+
+
+
+interface TrackClassification {
```
hardiesoft: We seem to be declaring `TrackClassification` twice?

types/api/fileProcessing.d.ts
```diff
@@ -90,6 +93,65 @@ export interface ThumbnailInfo {
...
+  message?: string;
+  model_used?: string;
+  raw_tag?: string;
+
+}
+
+export type MinimalTrackClassifications  =  MinimalTrackClassification[];
+
+
+export interface MinimalTrackClassification {
```
hardiesoft: Is `TrackClassification` supposted to be a superset of `MinimalTrackClassification`?  If so, could one type extend the other?

hardiesoft (changes requested):
Mostly looks pretty good, but there are a handful of things I think we can improve on, thanks!

api/api/V1/Processing.ts
```diff
@@ -387,39 +403,195 @@ export default function (app: Application, baseUrl: string) {
...
+      endSeconds: trackData.end_s,
+      minFreqHz: null,
+      maxFreqHz: null,
+      RecordingId: recording.id,
+    };
+    if (recording.type === RecordingType.Audio) {
+      newTrack.minFreqHz = trackData.minFreq || 0;
+      newTrack.maxFreqHz = trackData.maxFreq || 0;
+    }
+    if (trackData.start_s) {
```
hardiesoft: Prefer some other way to check the existence of this field: As written this will be false if `start_s` is 0.

hardiesoft (changes requested):
One m ore small fix, then I think this is looking pretty good now.

api/api/V1/Processing.ts
gferraro: i believe this field has  to exist so have removed the check

hardiesoft (approved):
Thanks for the changes! Looks good now, approved.

PR #127, opened 2026-05-27
fix failed state logic

gferraro:
use the same logic for failed state is non failed


hardiesoft (approved):
This looks reasonable, merging

